PR #44, opened 2020-08-16
Migrate tests to use compile testing

evant:
Blocked:
- ksp: needs to be updated
- kapt: not working on 1.4.0 https://github.com/tschuchortdev/kotlin-compile-testing/issues/51


evant:
Fails due to https://github.com/android/kotlin/issues/116

PR #51, opened 2020-08-29
Dynamic scoped modules

evant:


integration-tests-kapt/src/jmh/java/me/tatarka/inject/benchmark/ScopedAccessBenchmark.kt
```diff
@@ -0,0 +1,49 @@
...
+        get() = _foo
+}
+
+class InjectDynamicScopedAccessComponent : ScopedAccessComponent() {
+    private val _scope = ConcurrentHashMap<String, Any?>()
+
+    override val foo: Foo = _scope.lazyGet("me.tatarka.inject.benchmark.Foo") { Foo() }
+}
+
+open class ScopedAccessBenchmark {
```
evant, line 32: Comparing performance of field vs map impls. Seem to be the same order of magnitude.

kotlin-inject-runtime/src/main/kotlin/me/tatarka/inject/internal/LazyGet.kt
```diff
@@ -0,0 +1,16 @@
+package me.tatarka.inject.internal
+
+import java.util.concurrent.ConcurrentMap
+
+fun <T> ConcurrentMap<String, Any?>.lazyGet(key: String, init: () -> T): T {
```
evant, line 5: This is unfortunately java-specific, will have to figure out other platforms when that is possible. js is easy since there's no multithreading, not sure what to do about native.

Issue #10, opened 2020-07-08
Support annotating constructor with @Inject


evant:
Annotating the class selects the primary constructor which is usually what you want, but sometimes you'd want to annotate the constructor explicitly. This is useful if:
1. You have more than one constructor
2. You are migrating dagger code

PR #83, opened 2020-10-18
Support annotating constructors (primary and secondary) with `@Inject`

andrey-bolduzev:
Fixes https://github.com/evant/kotlin-inject/issues/10.

### DIFF
--- a/README.md
+++ b/README.md
@@ -362,8 +362,8 @@ You can provide some additional options to the processor.
 - `me.tatarka.inject.generateCompanionExtensions=true`
 
   This will generate the `create()` methods on the companion object instead of the component's class. This allows you to do
-  `MyComponent.create()` instead of `MyComponent::class.create()`. However, due to a kotlin limitiation you will have to
-  explitilcy specify a companion object for your component.
+  `MyComponent.create()` instead of `MyComponent::class.create()`. However, due to a kotlin limitation you will have to
+  explicitly specify a companion object for your component.
   
   ```kotlin
   @Component abstract class MyComponent {
--- a/kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/Ast.kt
+++ b/kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/Ast.kt
@@ -1,8 +1,6 @@
 package me.tatarka.inject.compiler
 
 import com.squareup.kotlinpoet.*
-import kotlin.contracts.InvocationKind
-import kotlin.contracts.contract
 import kotlin.reflect.KClass
 
 interface AstProvider {
@@ -57,6 +55,8 @@ abstract class AstClass : AstElement(), AstHasModifiers {
     abstract val superTypes: List<AstClass>
 
     abstract val primaryConstructor: AstConstructor?
+
+    abstract val constructors: List<AstConstructor>
 
     abstract val methods: List<AstMethod>
 
--- a/kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
+++ b/kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
@@ -628,13 +628,28 @@ fun AstElement.isComponent() = hasAnnotation<Component>()
 
 fun AstMethod.isProvides(): Boolean = hasAnnotation<Provides>()
 
-fun AstClass.isInject(options: Options): Boolean {
-    if (options.enableJavaxAnnotations) {
-        if (primaryConstructor?.hasAnnotation("javax.inject.Inject") == true) {
-            return true
+fun AstClass.findInjectConstructors(messenger: Messenger, options: Options): AstConstructor? {
+    val injectCtors = constructors.filter {
+        if (options.enableJavaxAnnotations) {
+            it.hasAnnotation("javax.inject.Inject") || it.hasAnnotation<Inject>()
+        } else {
+            it.hasAnnotation<Inject>()
         }
     }
-    return hasAnnotation<Inject>()
+
+    return when {
+        hasAnnotation<Inject>() && injectCtors.isNotEmpty() -> {
+            messenger.error("Cannot annotate constructor with @Inject in an @Inject-annotated class", this)
+            null
+        }
+        hasAnnotation<Inject>() -> primaryConstructor
+        injectCtors.size > 1 -> {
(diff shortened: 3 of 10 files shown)


kotlin-inject-compiler-kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/ModelAst.kt
```diff
@@ -243,6 +244,18 @@ private class ModelAstClass(
...
         }
 
+    override val constructors: List<AstConstructor>?
+        get() {
+            return ElementFilter.constructorsIn(element.enclosedElements).mapIndexed { index, constructor ->
+                ModelAstConstructor(
+                    this,
+                    this,
+                    constructor,
+                    kmClass?.constructors?.get(index)
```
andrey-bolduzev, line 254: This assumption that the order of constructors is the same from both sources is notably shaky. I assume this problem is similar to what the TODO above refers to as well. My low familiarity with the used APIs hasn't allowed me to solve this properly yet. The tests seem to be passing though.

evant: Yep you are correct. I believe this can be done similar to how methods are matched. Look up the kotlin metadata, then use it's jvm signature to match the corresponding element. Would have to take a closer look to make sure.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -628,13 +628,28 @@ fun AstElement.isComponent() = hasAnnotation<Component>()
...
+        if (options.enableJavaxAnnotations) {
+            it.hasAnnotation("javax.inject.Inject") || it.hasAnnotation<Inject>()
+        } else {
+            it.hasAnnotation<Inject>()
+        }
+    }.orEmpty()
+
+    return when {
+        hasAnnotation<Inject>() && injectCtors.isNotEmpty() -> {
+            messenger.error("Annotating constructors with @Inject in @Inject-annotated classes is not supported.", this)
```
evant, line 642: Would like to match the wording of other error messages so something like:
```suggestion
            messenger.error("Cannot annotate constructor with @Inject in an @Inject-annoated class", this)
```

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -628,13 +628,28 @@ fun AstElement.isComponent() = hasAnnotation<Component>()
...
+            null
+        }
+        hasAnnotation<Inject>() -> primaryConstructor
+        injectCtors.size > 1 -> {
+            messenger.error("Multiple @Inject-annotated constructors in the same class are not supported.", this)
```
evant, line 647: Ditto:
```suggestion
            messenger.error("Class cannot contain multiple @Inject-annotated constructors", this)
```

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/Ast.kt
```diff
@@ -58,6 +58,8 @@ abstract class AstClass : AstElement(), AstHasModifiers {
 
     abstract val primaryConstructor: AstConstructor?
 
+    abstract val constructors: List<AstConstructor>?
```
evant, line 61: Is there a reason this is nullable? I'd expect no constructors to be represented as an empty list.

kotlin-inject-compiler-kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/ModelAst.kt
andrey-bolduzev: Thank you, will look into it.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/Ast.kt
andrey-bolduzev: No reason. For some reason, I had an idea that it was the library preference (vs empty collections) but I now I can't find anything confirming that. Will switch over. Thank you.

kotlin-inject-compiler-kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/ModelAst.kt
andrey-bolduzev: Updated with your recommendation.

README.md
```diff
@@ -362,8 +362,8 @@ You can provide some additional options to the processor.
 - `me.tatarka.inject.generateCompanionExtensions=true`
 
   This will generate the `create()` methods on the companion object instead of the component's class. This allows you to do
-  `MyComponent.create()` instead of `MyComponent::class.create()`. However, due to a kotlin limitiation you will have to
-  explitilcy specify a companion object for your component.
+  `MyComponent.create()` instead of `MyComponent::class.create()`. However, due to a kotlin limitation you will have to
+  explicitly specify a companion object for your component.
```
evant, line 366: 👍

evant (changes requested):
Suggested a couple of minor modifications but other than that looks good!

kotlin-inject-compiler-kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/ModelAst.kt
```diff
@@ -232,20 +231,34 @@ private class ModelAstClass(
 
     override val primaryConstructor: AstConstructor?
         get() {
-            return ElementFilter.constructorsIn(element.enclosedElements).mapNotNull { constructor ->
-                //TODO: not sure how to match constructors
+            kmClass ?: return null
```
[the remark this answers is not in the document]
andrey-bolduzev: On a side note, have you considered adding ktlint and/or detekt and running checks and tests on each pull request? Not that it would catch this particular instance, ktlint doesn't have an opinion on this specific language construct, but it could save some time with future reviews and provide an opinionated code style to follow.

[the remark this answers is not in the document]
evant: 👍 Yeah, I should do that.

evant (approved):

PR #85, opened 2020-10-21
Add detekt and ktlint static checks

andrey-bolduzev:
Fixes https://github.com/evant/kotlin-inject/issues/84.

The ktlint config is kept to the absolute minimum and the code is mostly reformatted with `ktlint -F`.

The detekt config is a slightly stricter version of the default config, customized for the balance between how many `@Suppress` annotations are required in the code and the desire to keep important checks in.

@evant is enabling CI runs for PRs an option for us? Not sure how much room the current CircleCI plan leaves us but it would be great to run the checks and the tests on every push.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,8 +13,14 @@ jobs:
       - gradle/with_cache:
           steps:
             - run:
-                name: Run Tests
-                command: ./gradlew test
+                name: Run Tests and Checks
+                command: ./gradlew check
+            - run:
+                name: Collect Detekt Reports
+                command: cp -vr build/reports detekt
+                when: on_fail
+            - store_artifacts:
+                path: detekt
             - deploy:
                 command: |
                   if [ "${CIRCLE_BRANCH}" == "main" ]; then
--- a/.static/detekt-config.yml
+++ b/.static/detekt-config.yml
@@ -1,0 +1,167 @@
+build:
+  maxIssues: 0
+
+console-reports:
+  exclude:
+    - "ProjectStatisticsReport"
+    - "ComplexityReport"
+    - "NotificationReport"
+
+complexity:
+  LongParameterList:
+    active: false
+  LargeClass:
+    excludes: "**/test/**,**/integration-tests/**,**/*.Test.kt"
+  ReplaceSafeCallChainWithRun:
+    active: true
+  TooManyFunctions:
+    active: false
+
+formatting:
+  FinalNewline:
+    active: false
+  AnnotationSpacing:
+    active: true
+    autoCorrect: true
+
+naming:
+  ConstructorParameterNaming:
+    privateParameterPattern: "(_)?[a-z][A-Za-z0-9]*"
+  TopLevelPropertyNaming:
+    active: false
+  VariableNaming:
+    active: true
+    variablePattern: "[a-z][A-Za-z0-9]*"
+    privateVariablePattern: "(_)?[a-z][A-Za-z0-9]*"
+    excludeClassPattern: "$^"
+  NonBooleanPropertyPrefixedWithIs:
+    active: true
+    excludes: "**/test/**,**/integration-tests/**,**/*.Test.kt"
+  FunctionNaming:
+    ignoreAnnotated: ["Benchmark"]
+
+performance:
+  SpreadOperator:
+    active: false
+
+potential-bugs:
+  EqualsAlwaysReturnsTrueOrFalse:
+    active: true
+  InvalidRange:
+    active: true
+  IteratorHasNextCallsNextMethod:
+    active: true
+  IteratorNotThrowingNoSuchElementException:
+    active: true
+  UnconditionalJumpStatementInLoop:
+    active: true
+  UselessPostfixExpression:
+    active: true
+  WrongEqualsTypeParameter:
+    active: true
+  Deprecation:
+    active: true
+  HasPlatformType:
+    active: true
+  MissingWhenCase:
+    active: true
+  NullableToStringCall:
+    active: true
+  RedundantElseInWhen:
+    active: true
+  UnnecessaryNotNullOperator:
+    active: true
+  UnnecessarySafeCall:
+    active: true
+
+style:
+  ReturnCount:
+    active: false
+  LoopWithTooManyJumpStatements:
+    active: true
+  ForbiddenComment:
+    active: false
+  MagicNumber:
+    ignoreEnums: true
+    ignorePropertyDeclaration: true
+  MayBeConst:
+    active: true
+  NewLineAtEndOfFile:
+    active: false
+  RedundantVisibilityModifierRule:
+    active: true
+  SpacingBetweenPackageAndImports:
+    active: true
+  TrailingWhitespace:
+    active: true
+  UnnecessaryAbstractClass:
+    active: true
+    excludeAnnotatedClasses: ["Component"]
+    excludes: "**/test/**,**/integration-tests/**,**/*.Test.kt"
+  UnnecessaryApply:
+    active: true
+  UnnecessaryInheritance:
+    active: true
+  UnnecessaryParentheses:
+    active: true
+  UntilInsteadOfRangeTo:
+    active: true
+  UnusedImports:
+    active: true
+  UnusedPrivateMember:
+    active: true
+    excludes: "**/test/**,**/integration-tests/**,**/*.Test.kt"
+  UtilityClassWithPublicConstructor:
+    active: true
+    excludes: "**/test/**,**/integration-tests/**,**/*.Test.kt"
+  CollapsibleIfStatements:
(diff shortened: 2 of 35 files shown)


.circleci/config.yml
```diff
@@ -10,8 +10,23 @@ jobs:
...
+          name: Run Ktlint Checks
+          command: |
+            curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.39.0/ktlint
+            chmod a+x ktlint
+            ./ktlint
       - gradle/with_cache:
           steps:
+            - run:
+                name: Run Detekt Checks
+                command: ./gradlew detektAll
```
evant, line 23: I'm pretty sure you can run `./gradlew check` and it'll run detekt and tests.

.static/detekt-config.yml
```diff
@@ -0,0 +1,172 @@
...
+    excludeClassPattern: "$^"
+  NonBooleanPropertyPrefixedWithIs:
+    active: true
+    excludes: "**/test/**,**/*.Test.kt"
+  FunctionNaming:
+    ignoreAnnotated: ["Benchmark"]
+
+performance:
+  SpreadOperator:
+    active: false
```
evant, line 54: Just saw a potential fix for this https://github.com/detekt/detekt/releases/tag/v1.14.2 so maybe it doesn't need to be disabled?

build.gradle
```diff
@@ -1,11 +1,33 @@
...
     }
+}
+
+task detektAll(type: io.gitlab.arturbosch.detekt.Detekt) {
+    parallel = true
+    failFast = false
+    buildUponDefaultConfig = true
+    source = files(projectDir)
+    config.from = project.file(".static/detekt-config.yml")
+    include("**/*.kt")
```
evant, line 24: are these include/excludes needed? I don't remember needing them.

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/AbstractComponentTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("UnnecessaryAbstractClass")
```
evant, line 1: Might be worth having separate main and test configs, have the test code be more lenient since it seems to incorrectly be flagging test cases.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -498,6 +499,7 @@ class InjectGenerator(provider: AstProvider, private val options: Options) :
         return find(key)
     }
 
+    @Suppress("ReturnCount")
```
evant, line 502: Not a fan of this check, I love early returns. Should be disabled.

.circleci/config.yml
```diff
@@ -10,8 +10,23 @@ jobs:
       JVM_OPTS: -Xmx3200m
     steps:
       - checkout
+      - run:
+          name: Run Ktlint Checks
+          command: |
+            curl -sSLO https://github.com/pinterest/ktlint/releases/download/0.39.0/ktlint
+            chmod a+x ktlint
+            ./ktlint
```
evant, line 18: Doesn't detekt have ktlint integration? I wouldn't expect to have to download it separately.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -11,6 +19,7 @@ import kotlin.reflect.KClass
 private val SCOPED_COMPONENT = ClassName("me.tatarka.inject.internal", "ScopedComponent")
 private val LAZY_MAP = ClassName("me.tatarka.inject.internal", "LazyMap")
 
+@Suppress("TooManyFunctions")
```
evant, line 22: Should this one be disabled? I can't think of a case where I'd see it and _not_ suppress it.

evant:
> is enabling CI runs for PRs an option for us?

Hm, I did try to enable it after your last pr, need to figure out why it's not working.

.circleci/config.yml
andrey-bolduzev: Not in this configuration. It would work like this if the `detekt` plugin was applied to all submobules separately instead of only applying it to the top-level `build.gradle`. From personal experience, depending in the project size, running 20 `detekt` tasks is much slower then running the single `detektAll`. That and the overhead of having the plugin code in every `build.gradle` in the project.

It's still possible to make `check` depend on the custom `detektAll` task, of course, if we  want to. I'm not aware of any deep integration between detekt and ktlint (apart from sharing some rules) but running ktlint can be delegated to gradle as well. In which case I can make `check` depend on that task as well.

build.gradle
andrey-bolduzev: Yep, you're right.

evant:
Think if figured out the ci building thing https://github.com/evant/kotlin-inject/pull/86

.static/detekt-config.yml
andrey-bolduzev: This check is not broken, it's just, personally, I don't know what I would do about it. I like spreading arrays and I'm willing to pay the price.

In the PR branch, for instance, it complains about this snippet (`ModelAst.kt:135`):
```kotlin
    private fun elementDeclaredType(type: TypeElement, astTypes: List<AstType>) =
        types.getDeclaredType(
            type,
            *astTypes.map { // this line
                (it as ModelAstType).type
            }.toTypedArray()
        )
```

.circleci/config.yml
evant: You can add ktlint as a detekt plugin https://detekt.github.io/detekt/#adding-more-rule-sets

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
andrey-bolduzev: :+1:

build.gradle
evant: heh, yeah

andrey-bolduzev: Actually, sorry, I think the `include` is still needed, otherwise, stuff like this is outputted:
```
...
Ignoring a file detekt cannot handle: .../kotlin-inject/integration-tests/module/build.gradle
Ignoring a file detekt cannot handle: .../kotlin-inject/integration-tests/ksp-companion/build.gradle
Ignoring a file detekt cannot handle: .../kotlin-inject/integration-tests/kapt/build.gradle
Ignoring a file detekt cannot handle: .../kotlin-inject/integration-tests/kapt-companion/build.gradle
Ignoring a file detekt cannot handle: .../kotlin-inject/LICENSE
...
```

evant: Looking at my other project, you can set the source to the sourceSet explicitly, that should be cleaner than wildcards imo https://github.com/willowtreeapps/assertk/blob/cf0d1da83c71130a12f81b0665d4faf68fc2eb1b/build.gradle#L72

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
andrey-bolduzev: I suppose there are different schools of thought on what makes code readable and I'm not here to preach :) Disabling.

.static/detekt-config.yml
evant: Ah yep, the linked fixes doesn't actually solve it.

build.gradle
andrey-bolduzev: Thanks for the reference. It does look interesting but I have to note that that approach has a slight inconvenience of needing to update the detekt task config when a new module/source set is added if I understand correctly. Or more like needing to _not forget_ to do that.

andrey-bolduzev:
> Think I figured out the ci building thing #86

Nice, seems to be working.

.circleci/config.yml
andrey-bolduzev: Thank you, didn't know about this ktlint integration. All folded into `check` now to the best of my groovy-writing abilities.

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/AbstractComponentTest.kt
andrey-bolduzev: Upon closer look, there were only 2 checks that the tests were triggering, so I went with a lighter-weight approach of excluding the test files in a few checks. The extra `@Suppress` sites are now gone. Tell me if this works.

evant (approved):

Issue #59, opened 2020-09-17
Support injecting objects


evant:
```kotlin
@Inject object Foo
```

can be useful if you want to be able to change it from a singleton to an instance later.

PR #87, opened 2020-10-24
Support injecting objects

japplin:
Fixes https://github.com/evant/kotlin-inject/issues/59

### DIFF
--- a/integration-tests/common/src/test/kotlin/me/tatarka/inject/test/InjectObjectTest.kt
+++ b/integration-tests/common/src/test/kotlin/me/tatarka/inject/test/InjectObjectTest.kt
@@ -1,0 +1,41 @@
+package me.tatarka.inject.test
+
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import me.tatarka.inject.annotations.Component
+import me.tatarka.inject.annotations.Inject
+import me.tatarka.inject.test.different.DifferentPackageFoo
+import org.junit.Test
+
+@Inject object MyInjectObject
+
+@Inject class DependsOnMyInjectObject(val myInjectObject: MyInjectObject)
+
+@Component interface MyComponent {
+    val injectObject: MyInjectObject
+    val differentPackageObject: DifferentPackageFoo.MyObject
+    fun injectObject(): MyInjectObject
+    fun dependsOnMyInjectObject(): DependsOnMyInjectObject
+}
+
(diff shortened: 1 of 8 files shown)


kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
```diff
@@ -247,6 +247,9 @@ class TypeCollector private constructor(private val provider: AstProvider, priva
                 scopedComponent = scopedComponent?.type
             )
         }
+        if (astClass.isObject) {
```
andrey-bolduzev, line 250: I'm thinking what if we move this to be above the `findInjectConstructors` call and remove the check inside the `findInjectConstructors`?

japplin: This does work but it then findInjectConstructors returns a result for object classes which is incorrect since it generates code which doesn't compile.

japplin: Remove check from here and added it to `ModelAstClass#primaryConstructor` to prevent it from incorrectly returning constructors for object classes.

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/InjectObjectTest.kt
```diff
@@ -0,0 +1,41 @@
...
+    val injectObject: MyInjectObject
+    val differentPackageObject: DifferentPackageFoo.MyObject
+    fun injectObject(): MyInjectObject
+    fun dependsOnMyInjectObject(): DependsOnMyInjectObject
+}
+
+class InjectObjectTest {
+    @Test
+    fun inject_annotated_object_can_be_provided_in_component_function() {
+        assertThat(InjectMyComponent().injectObject(), equalTo(MyInjectObject))
```
evant, line 24: Please use assertk assertions to match the rest of the tests, ex: `assertThat(InjectMyComponent().injectObject()).isEqualTo(MyInjectObject)`

evant (changes requested):
Other than the tests, looks good!

evant (approved):

Issue #88, opened 2020-10-24
Generated class for internal abstract class should also be internal


japplin:

evant:
Hm yeah will have to think about visibility. I image you mean the generated create method should be marked internal? The class shouldn't be accessed directly and it should probably have a smaller visibility regardless. Will have to make sure this works correctly with parent-child components though.

PR #89, opened 2020-10-25
Respect visibility when generating code

japplin:
Fixes https://github.com/evant/kotlin-inject/issues/88
I imagine this is probably missing some cases still just wanted to get this up to get others opinions

### DIFF
--- a/integration-tests/common/src/test/kotlin/me/tatarka/inject/test/VisibilityTest.kt
+++ b/integration-tests/common/src/test/kotlin/me/tatarka/inject/test/VisibilityTest.kt
@@ -1,0 +1,64 @@
+package me.tatarka.inject.test
+
+import assertk.assertThat
+import assertk.assertions.isEqualTo
+import me.tatarka.inject.annotations.Component
+import org.junit.Test
+import kotlin.reflect.KVisibility
+import kotlin.reflect.full.functions
+import kotlin.reflect.full.memberProperties
+
+@Component
+abstract class VisibilityTestComponent {
+    abstract val myPublicProperty: Foo
+    internal abstract val myInternalProperty: InternalBarConstructor
+    protected abstract val myProtectedProperty: Foo
+
+    abstract fun myPublicFunction(): InternalBarConstructor
+    internal abstract fun myInternalFunction(): InternalFoo
+    protected abstract fun myProtectedFunction(): Foo
+}
+
+@Component internal interface InternalVisibilityTestComponent
+
+class VisibilityTest {
+
+    @Test
+    fun public_component_generates_public_properties() {
+        assertThat(property("myPublicProperty").visibility).isEqualTo(KVisibility.PUBLIC)
+    }
+
+    @Test
+    fun public_component_generates_internal_properties() {
+        assertThat(property("myInternalProperty").visibility).isEqualTo(KVisibility.INTERNAL)
+    }
+
+    @Test
+    fun public_component_generates_protected_properties() {
+        assertThat(property("myProtectedProperty").visibility).isEqualTo(KVisibility.PROTECTED)
+    }
+
+    @Test
+    fun public_component_generates_public_function() {
+        assertThat(function("myPublicFunction").visibility).isEqualTo(KVisibility.PUBLIC)
+    }
+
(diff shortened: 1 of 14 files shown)


japplin:
@evant Any thoughts on this change?

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/VisibilityTest.kt
```diff
@@ -0,0 +1,64 @@
...
+import assertk.assertions.isEqualTo
+import me.tatarka.inject.annotations.Component
+import org.junit.Test
+import kotlin.reflect.KVisibility
+import kotlin.reflect.full.functions
+import kotlin.reflect.full.memberProperties
+
+@Component
+abstract class VisibilityTestComponent {
+    abstract val myPublicProperty: Foo
```
evant, line 13: Screen Shot 2020-11-08 at 5 14 26 PM
It looks like when overriding you don't need to specify the modifier.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/AstVisibility.kt
```diff
@@ -0,0 +1,16 @@
+package me.tatarka.inject.compiler
+
+import com.squareup.kotlinpoet.KModifier
+
+enum class AstVisibility { PUBLIC, PRIVATE, PROTECTED, INTERNAL }
```
evant, line 5: I actually had started with a modifier enum but switched to methods like `isPrivate` because it felt simpler and matched the ksp api.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/CreateGenerator.kt
```diff
@@ -0,0 +1,80 @@
...
+
+import com.squareup.kotlinpoet.CodeBlock
+import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.plusParameter
+import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.asClassName
+import kotlin.reflect.KClass
+
+class CreateGenerator(private val astProvider: AstProvider, private val options: Options) {
```
evant, line 11: I'm a bit skeptical that this needs to be broken up into it's own class. And if so, it should be an implementation detail of `InjectGenerator`, instead of being passed as an argument.

evant (changes requested):
Sorry for the delay. Finally got a chance to properly look at this.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/AstVisibility.kt
japplin: I think having a state which represents which visibility an element well since it can only be 1 of these states at a time, and never more than one which this enum makes clear. Perhaps adding methods to the Visibility interface to check if it's public, private, internal or protected would help?

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/CreateGenerator.kt
japplin: I think this class nicely groups together the code for generating a components create method without the need for functions defined in functions. I didn't want to do too many changes at once in this pr but I also think it would be possible to remove the Options flag from this class and instead have several impls of a `CreateGenerator` interface `CompanionCreateGenerator` & `ClassCreateGenerator`.

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/VisibilityTest.kt
japplin: I'll remove the visibility modifier from properties when they are generated

evant:
Made some modifications and merged

Issue #56, opened 2020-09-17
Generate getters for common object construction


evant:
Right now the entire constructor/provides call is generated every place a dependency is used. This leads to a bit of code bloat. Should be able to pull up the construction into a private getter if it's used more than once in a component.

ex:
```kotlin
class InjectMyComponent : MyComponent() {
  private val _foo: Foo
    get() = Foo(...)

  override val bar1: Bar1
    get() = Bar1(_foo)

  override val bar2: Bar2
    get() = Bar2(_foo)
```

PR #100, opened 2021-04-06
General mechanism to optimize duplicate code

evant:
We construct the dependency graph re-using TypeResults when we can. We then
iterate over the graph and insert private getters above nodes which have
children and more than one parent.

Fixes https://github.com/evant/kotlin-inject/issues/56


kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
```diff
@@ -0,0 +1,178 @@
...
+class TypeResultResolver(private val options: Options) {
+
+    private val types = mutableMapOf<TypeKey, TypeResult>()
+
+    /**
+     * Find the given type in this context.
+     */
+    fun find(context: Context, key: TypeKey): TypeResultRef {
+        return TypeResultRef(key, types[key] ?: context.findType(key).linkRefs()
+            .also { types[key] = it })
```
nishtahir, line 18: Might be worth renaming things here to reflect the memorization happening here

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    private fun Context.findType(key: TypeKey): TypeResult {
+        val typeCreator = collector.resolve(key)
+        if (typeCreator != null) {
+            return typeCreator.toResult(this, key, skipScoped)
+        }
+        if (key.type.isFunction()) {
+            val resolveType = key.type.resolvedType()
+            if (key.type.isTypeAlis()) {
```
nishtahir, line 45: Nit: `isTypeAlis` should be `isTypeAlias`?

nishtahir (approved):
LGTM, some feedback

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultTreeOptimizer.kt
```diff
@@ -0,0 +1,100 @@
...
+    val returnType: AstType,
+    val isProperty: Boolean,
+    val isPrivate: Boolean,
+    val isOverride: Boolean,
+    val isSuspend: Boolean,
+    val typeResult: TypeResultRef,
+)
+
+@Suppress("NestedBlockDepth")
+fun List<MethodEntry>.optimize(): List<MethodEntry> {
```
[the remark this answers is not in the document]
evant: Feels kinda yagni to me, can always make it a class in the future if necessary.

PR #103, opened 2021-04-08
Update cycle detection to allow cycles when construction is delayed.

### LINKED ISSUE #19: Allow recusive dependencies with lazy/functions
Right now this stackoverflows but it should be able to be resolved:

```kotlin
class LazyCycleFoo(val bar: LazyCycleBar)
class LazyCycleBar(val foo: Lazy<LazyCycleFoo>)

@Component abstract class LazyCycleComponent {
    abstract val bar: LazyCycleBar

    @Provides fun bar(foo: Lazy<LazyCycleFoo>) = LazyCycleBar(foo)

    @Provides fun foo(bar: LazyCycleBar) = LazyCycleFoo(bar)
}
```

should generate:
```kotlin
class InjectLazyCycleComponent : LazyCycleComponent() {
    override val bar: LazyCycleBar
        get() {
            lateinit var bar: LazyCycleBar
            return bar(lazy { foo(bar) }).also { bar = it }
        }
}
```

evant:
You can break a cycle with Lazy or a function. The generated code will use
lateinit to be able to pass the variable before it's constructed. It is a runtime
exception to try to access the dependency before construction completes.

Fixes #19


evant:
Note: found a pretty big bug with this. It breaks the optimizer because it might extract out a getter that references the local lateinit var.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/CycleDetector.kt
```diff
@@ -2,25 +2,57 @@ package me.tatarka.inject.compiler
...
+            } else {
+                CycleResult.Cycle
+            }
+        } else {
+            _elements.add(Entry.Element(element))
             CycleResult.None
         }
     }
 
     fun pop() {
```
nishtahir, line 42: Not sure if this is a good name for it anymore. Since it may pop more than one thing

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/CycleDetector.kt
```diff
@@ -2,25 +2,57 @@ package me.tatarka.inject.compiler
 
 class CycleDetector {
 
-    private val _elements = mutableListOf<AstElement>()
+    private val _elements = mutableListOf<Entry>()
+    private var resolving: CycleResult.Resolvable? = null
```
nishtahir, line 6: Can there be more than one lazy element resolved at a time?

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -232,4 +230,32 @@ fun AstMethod.isProvider(): Boolean =
     !hasAnnotation<Provides>() && isAbstract && when (this) {
         is AstFunction -> parameters.isEmpty()
         is AstProperty -> true
-    } && receiverParameterType == null && returnType.isNotUnit()
\ No newline at end of file
+    } && receiverParameterType == null && returnType.isNotUnit()
+
+private fun dumpGraph(astClass: AstClass, entries: List<MethodEntry>): String {
```
nishtahir, line 235: Maybe move this into its own file? Seems like pretty generic logic

nishtahir (approved):

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/CycleDetector.kt
evant: Good question, will try to figure out if I can write a test where that is the case.

evant: Yep, added a failing test-case and fixed it.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
evant: Kinda meh about that as it's only used here.

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/CycleDetector.kt
evant: Cleaned up so the pop method is no longer needed.

PR #104, opened 2021-04-12
Cycle fixes

evant:
Fixes issues with late-init references being cached & reused across method boundaries. I don't know if I love this solution but it works.


kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
```diff
@@ -235,4 +259,19 @@ class TypeResultResolver(private val provider: AstProvider, private val options:
      * elements that were traversed for this context.
      */
     private fun trace(message: String): String = "$message\n" + cycleDetector.trace(provider)
+
+    private val TypeResult.isCacheable: Boolean
```
evant, line 263: This feels n^2 which I'm not loving, but maybe it's ok?

PR #109, opened 2021-04-15
Correctly handle components declared as inner classes

### LINKED ISSUE #107: Namespace generated inner classes
If you have two components with the same name with different parent classes, they generate conflicting classes.

```kotlin
class Foo1 {
  @Component abstract class MyComponent()
}
class Foo2 {
  @Component abstract class MyComponent()
}
```

both generates `InjectMyComponent()` which causes a compile error.

evant:
Fixes #107

### DIFF
--- a/kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
+++ b/kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
@@ -43,11 +43,12 @@ class InjectGenerator<Output, Provider>(
 
         val constructor = astClass.primaryConstructor
 
-        val injectComponent = generateInjectComponent(astClass, constructor)
+        val injectName = astClass.toInjectName()
+        val injectComponent = generateInjectComponent(astClass, injectName, constructor)
         val createFunction = createGenerator.create(astClass, constructor, injectComponent.typeSpec)
 
         return provider.astFileSpec(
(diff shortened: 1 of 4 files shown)


kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -214,6 +217,9 @@ fun AstMethod.isProvider(): Boolean =
         is AstProperty -> true
     } && receiverParameterType == null && returnType.isNotUnit()
 
+fun AstClass.toInjectName(): String =
+    "Inject${asClassName().simpleNames.joinToString("_")}"
```
nishtahir, line 221: How does this work with class names with backticks?

nishtahir (approved):

kotlin-inject-compiler-core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
evant: Backticks work in variable names too so I think it still works?

evant: Or maybe not because you'd get 'Inject\`My backtick class`'

evant: https://github.com/evant/kotlin-inject/issues/113

Issue #110, opened 2021-04-15
Properly handle nullable/platform types.


evant:
Kotlin has 3 versions of a type: non-nullable, nullable, and platform, represented as:
`Foo`, `Foo?`, and `Foo!` respectively. We need to handle these correctly and consistently across backends. The current behavior is:

- kapt: all 3 types match with each other. This 'works' sometimes but generates incorrect code when it matches `Foo?` to `Foo`.
- ksp: all 3 types are distinct. This works, but can be surprising in the case of platform times where `@Provides fun foo() = javaMethodThatReturnsFoo()` won't match `Foo`.

To say another way, which ones of these should compile, and which ones should complain about a missing binding?

```kotlin
@Component MyComponent {
  abstract val foo: Foo
  @Provides fun foo() = javaMethodThatReturnsFoo()
}
```

```kotlin
@Component MyComponent {
  abstract val foo: Foo?
  @Provides fun foo() = javaMethodThatReturnsFoo()
}
```

```kotlin
@Component MyComponent {
  abstract val foo: Foo?
  @Provides fun foo(): Foo = javaMethodThatReturnsFoo()
}
```

evant:
Got a couple of thought here, we could either:
1) Be strict like ksp is (I think the metadata is there in the kapt backend to do this)
2) Allow types to be 'upgraded' when compatible. This would be similar to how you can do both `foo: Foo = javaMethodThatReturnsFoo()` and `foo: Foo? = javaMethodThatReturnsFoo()` in kotlin.

I'm leaning towards 1 for fewer surprises but it does mean you may have to explicitly declare the return type on a `@Provides` in some cases. Maybe an explicit warning/error message here could help?

sunian:
I'd prefer option 1 as well.

rf43:
I agree with @sunian that as a consumer, option 1 seems to be the safest, most forward looking solution.

Also, `may have to explicitly declare the return type on a @Provides in some cases` while slightly inconvenient I wouldn't think that's too bad.

evant:
It's not inconvenient no, but it did take me a bit to figure out what was happening. I had
```kotlin
@Provides fun retrofit(client: OkHttpClient) = Retrofit.Builder().build()
@Provides fun myService(retrofit: Retrofit): MyService = retrofit.create()
```
and it was complaining that it couldn't find retrofit. Again a targeted warning/error might be good enough though.

evant:
[...]

PR #111, opened 2021-04-15
Compare nullable & non-nullable types as different on kapt

evant:
This fixes the most egregious issue with #110


nishtahir (approved):
LGTM

PR #112, opened 2021-04-15
Improve ksp performance by reducing what we resolve

evant:


kotlin-inject-compiler-ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/KSAst.kt
```diff
@@ -324,6 +329,11 @@ private class KSAstProperty(provider: KSAstProvider, override val declaration: K
 
     override fun returnTypeFor(enclosingClass: AstClass): AstType {
         require(enclosingClass is KSAstClass)
+        // resolver.asMemberOf is expensive and only matters if there are generics involved, so check that first.
```
nishtahir, line 332: Nice

nishtahir (approved):

PR #117, opened 2021-04-19
Add android docs

evant:
[Rendered](https://github.com/evant/kotlin-inject/blob/e01245ea973d4479c4ace71a3ffda56afaf8c47b/docs/android.md)


nishtahir (approved):
LGTM covers the what I think would be the FAQ

PR #118, opened 2021-04-19
Look through type alias for @IntoMap

### LINKED ISSUE #105: Support typealias for @IntoMap
Normally typealiases are treated as separate types, but here it should look through them to see the Pair as it's useful to re-use across provides.

```kotlin
typealias MyEntry = Pair<MyKey, MyValue>

@Component
abstract class MyComponent {
   @Provides @IntoMap
   fun entry1(): MyEntry = ...
   @Provides @IntoMap
   fun entry2(): MyEntry = ...
}
```

evant:
Fixes #105

### DIFF
--- a/kotlin-inject-compiler/kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/Util.kt
+++ b/kotlin-inject-compiler/kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/Util.kt
@@ -301,6 +301,18 @@ fun KmType.eqvHashCode(collector: HashCollector = HashCollector()): Int {
     }
 }
 
+/**
+ * Returns the [KmType] without type alias info.
+ */
+fun KmType.resolve(): KmType = KmType(flags).also {
+    it.classifier = classifier
+    it.arguments.clear()
+    it.arguments.addAll(arguments)
+    it.abbreviatedType = null
+    it.outerType = outerType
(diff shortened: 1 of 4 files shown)


kotlin-inject-compiler/kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/Util.kt
```diff
@@ -301,6 +301,18 @@ fun KmType.eqvHashCode(collector: HashCollector = HashCollector()): Int {
     }
 }
 
+/**
+ * Returns the [KmType] without type alias info.
+ */
+fun KmType.resolve(): KmType = KmType(flags).also {
```
nishtahir, line 307: Maybe use `apply` here? At first glance this doesn't seem to be performing any side effects so I would avoid using also here

nishtahir (approved):

kotlin-inject-compiler/kapt/src/main/kotlin/me/tatarka/inject/compiler/kapt/Util.kt
evant: I did originally, but imo
```
classifier = classifier
```
is even more confusing

nishtahir: Makes sense. Probably worth adding a comment about. LGTM 👍

PR #128, opened 2021-05-13
Update to kotlin 1.5.0 & ksp alpha10

evant:


evant:
Note: the kotlinx-metadata update fixes https://github.com/evant/kotlin-inject/issues/130

PR #175, opened 2021-11-16
Fix conflicting declarations with scoped @Provides functions

### LINKED ISSUE #174: Conflicting declarations for scoped `@Provides` functions with parameterized return type
If I have:

```
import me.tatarka.inject.annotations.Provides

@Scope annotation class MyScope

@MyScope @Component abstract class MyComponent {
  abstract val foo: Foo
 
  @MyScope @Provides protected fun t1() = T<Int>()
 
  @MyScope @Provides protected fun t2() = T<String>()
}

class T<T>

@MyScope @Inject class Foo(
  private val t1: T<Int>,
  private val t2: T<String>,
  private val bar: Bar
)

@MyScope @Inject class Bar(
  private val t1: T<Int>,
  private val t2: T<String>
)
```

then the generated properties for both are named `_t`.

eygraber:
- If multiple scoped @Provides functions have the same return type with different type args, the generated properties will have the same name
  - This causes conflicting declaration errors while compiling
  - To fix that, the type args are included in property name generation

Fixes #174

### DIFF
--- a/integration-tests/common/src/test/kotlin/me/tatarka/inject/test/ScopeTest.kt
+++ b/integration-tests/common/src/test/kotlin/me/tatarka/inject/test/ScopeTest.kt
@@ -149,6 +149,37 @@ abstract class NestedExternalScopedComponent(@Component val parent: ExternalChil
     abstract val foo: IExternalFoo
 }
 
+class Parameterized<E, S>
+
+@CustomScope
+@Inject
+class ParameterizedFoo(
+  private val p1: Parameterized<List<Int>, List<String>>,
+  private val p2: Parameterized<List<String>, List<String>>,
+  private val bar: ParameterizedBar
+)
+
+@CustomScope
+@Inject
+class ParameterizedBar(
+  private val p1: Parameterized<List<Int>, List<String>>,
+  private val p2: Parameterized<List<String>, List<String>>
+)
+
+@CustomScope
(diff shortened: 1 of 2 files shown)


kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
```diff
@@ -263,4 +265,45 @@ class RoundsTest {
             doesNotContain("Cannot provide", "as it is already provided")
         }
     }
-}
\ No newline at end of file
+
+  @ParameterizedTest
+  @EnumSource(Target::class)
+  fun scoped_provides_funs_that_have_the_same_return_type_with_different_type_args_does_not_conflict(target: Target) {
```
evant, line 271: Since this is a happy-path test, it can go it integration-tests instead.

evant (commented):
Thanks! 1 comment about the test but other than that looks good!

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
eygraber: Updated

evant (approved):

PR #176, opened 2021-11-17
Generate pretty code

eygraber:


kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
```diff
@@ -115,33 +115,53 @@ data class TypeResultGenerator(val options: Options, val implicitAccessor: Acces
             if (isProperty) {
                 add("%N", methodName)
             } else {
-                add("%N(", methodName)
-                parameters.forEachIndexed { i, param ->
+                add("%N", methodName)
+                val firstParam = parameters.firstOrNull()?.result
+                if (firstParam is TypeResult.Function) {
```
evant, line 120: Hm, not sure I love how much this has to know about it's parameters, I wonder if there's a way to improve this?

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
```diff
@@ -254,17 +280,26 @@ data class TypeResultGenerator(val options: Options, val implicitAccessor: Acces
             if (args.isNotEmpty()) {
                 add(" ->")
             }
+            add("\n⇥")
 
-            add("%M(", name)
-            parameters.forEachIndexed { i, param ->
+            add("%M", name)
+            val firstParam = parameters.firstOrNull()?.result
+            if (firstParam is TypeResult.Function || firstParam is TypeResult.NamedFunction) {
```
evant, line 287: ditto, I wonder if there's a way to 'push' this logic into generating code for `TypeResult.Function`/`TypeResult.NamedFunction`.

evant (commented):
Generated code looks good! Have a couple of comments on maybe improving how it's implemented.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
eygraber: Hmm looking at it again, it seems there's a bug. The intention was to use trailing lambda syntax for functions that only have a single lambda parameter, but if there are multiple params and the first one is a lambda, then it will only write out the first param :scream: 

That bug aside, I thought it was much more readable that way, but lambdas in general get weird because they have newlines but other params don't. What are your thoughts on that?

eygraber: Looking at it more, I'm going to remove that part. It's the difference between seeing:

```
bar({
  foo(fBar)
})
```

and 

```
bar {
  foo(fBar)
}
```

which I think for generated code is fine.

eygraber: Updated

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
```diff
@@ -128,20 +128,29 @@ data class TypeResultGenerator(val options: Options, val implicitAccessor: Acces
...
             }
         }.build()
     }
 
     private fun TypeResult.Constructor.generate(): CodeBlock {
         return CodeBlock.builder().apply {
             add("%T(", type.asTypeName())
-            parameters.forEachIndexed { i, param ->
+            if (parameters.isNotEmpty()) {
+              val constructorParams = type.toAstClass().primaryConstructor?.parameters ?: emptyList()
```
evant, line 141: Think it would be better to add a `name` property to `TypeResult.Constructor`, then you don't have to re-look-it-up.

evant: To be more specific, you can change the type of `parameters` from `List<TypeResultRef>` to `Map<String, TypeResultRef>`

eygraber: I added a new commit with those changes. Because of that I was also able to provide named parameters to functions.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
```diff
@@ -116,32 +116,50 @@ data class TypeResultGenerator(val options: Options, val implicitAccessor: Acces
...
         }.build()
     }
 
     private fun TypeResult.Constructor.generate(): CodeBlock {
         return CodeBlock.builder().apply {
             add("%T(", type.asTypeName())
-            parameters.forEachIndexed { i, param ->
+            if (parameters.isNotEmpty()) {
+              add("\n⇥")
+              val isNamedArgumentsSupported = type.toAstClass().primaryConstructor != null
```
evant, line 149: Curious why you need this? And note: TypeResult.Constructor may not be the primary constructor as you can do:
```kotlin
class Foo() {
   @Inject constructor(bar: Bar) : this()
}
```

eygraber: Ah sorry, meant to leave a comment about that. That was the only way I could figure out to detect if this was a java class or not, because if it is a java class we can't use named parameters.

eygraber: I'd rather detect that some other way if it is possible. If not, I'll just leave a comment explaining why.

evant: Oh interesting, yeah this would have to be exposed through the Ast, kapt could check if metadata exists, and ksp should have a method to check if if it's a java class or not. This actually would be necessary as a kotlin class can actually have no primary constructor.

In terms of in the generator, I think it would make sense to have an additional `supportsNamedArguments` field on `TypeResult.Constructor` to make intentions clear and not have to call `toAstClass()` again.

If you don't want to dive that deep into the Ast, I don't mind taking things from here.

eygraber: I just pushed something like that. Let me know if there's a better way to do it.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/KSAst.kt
```diff
@@ -165,6 +165,8 @@ private class KSAstBasicElement(private val declaration: KSDeclaration) : AstBas
 
 private class KSAstClass(override val resolver: Resolver, override val declaration: KSClassDeclaration) :
     AstClass(), KSAstHasModifiers {
+    override val isJavaClass: Boolean
+        get() = declaration.findAnnotation("kotlin", "Metadata") == null
```
evant, line 169: I think you can use
```suggestion
        get() = declaration.origin == Origin.JAVA || declaration.origin == Origin.JAVA_LIB
```

eygraber: Updated

evant (approved):
Looks good!

Issue #167, opened 2021-11-10
Allow mutlithreading on kotlin native


evant:
Right now components can only be accessed on the main thread due to the `LazyMap` implementation. Need to figure out how to allow multithreading like what's available on the jvm. The solution should also work with kotlin's [new memory model](https://blog.jetbrains.com/kotlin/2021/08/try-the-new-kotlin-native-memory-manager-development-preview/).

saket:
I don't think this is fixed yet. I am running into `InvalidMutabilityException` error when running my project on kotlin/native:

```
Uncaught Kotlin exception: kotlin.native.concurrent.InvalidMutabilityException: mutation attempt of frozen kotlin.collections.HashMap@3ce1088
    at 0   Teleport                            0x0000000105a40c6c kfun:kotlin.Throwable#<init>(kotlin.String?){} + 96
    at 1   Teleport                            0x0000000105a39aa4 kfun:kotlin.Exception#<init>(kotlin.String?){} + 92
    at 2   Teleport                            0x0000000105a39cd8 kfun:kotlin.RuntimeException#<init>(kotlin.String?){} + 92
    at 3   Teleport                            0x0000000105a6de40 kfun:kotlin.native.concurrent.InvalidMutabilityException#<init>(kotlin.String){} + 92
    at 4   Teleport                            0x0000000105a6f254 ThrowInvalidMutabilityException + 336
    at 5   Teleport                            0x0000000105c09334 MutationCheck + 156
    at 6   Teleport                            0x0000000105a52568 kfun:kotlin.collections.HashMap.<set-length>#internal + 96
    at 7   Teleport                            0x0000000105a56920 kfun:kotlin.collections.HashMap#addKey(1:0){}kotlin.Int + 1048
    at 8   Teleport                            0x0000000105a535e8 kfun:kotlin.collections.HashMap#put(1:0;1:1){}1:1? + 236
    at 9   Teleport                            0x00000001060b9ebc kfun:me.tatarka.inject.internal.LazyMap#get(kotlin.String;kotlin.Function0<0:0>){0§<kotlin.Any?>}0:0 + 1008
    at 10  Teleport                            0x00000001057d31f4 kfun:me.saket.teleport.di.InjectAppComponent#<get-strings>(){}me.saket.teleport.localization.AppStrings + 232
[...]

PR #177, opened 2021-11-18
Annotate a top-level property in LazyMap with @SharedImmutable

saket:
Kotlin/native otherwise crashes when `NULL` is accessed:

```
Uncaught Kotlin exception: kotlin.native.IncorrectDereferenceException: Trying to access top level value not marked as @ThreadLocal or @SharedImmutable from non-main thread
```

I actually ran into this error when I tried my components using `IsoState` as per your suggested in https://github.com/evant/kotlin-inject/issues/167#issuecomment-970449637.

### DIFF
--- a/kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
+++ b/kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
@@ -1,8 +1,10 @@
 package me.tatarka.inject.internal
 
+import kotlin.native.concurrent.SharedImmutable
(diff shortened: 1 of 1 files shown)


evant (approved):

evant:
Might be a good idea to add to/update the test in https://github.com/evant/kotlin-inject/blob/main/kotlin-inject-runtime/src/nativeTest/kotlin/me/tatarka/inject/internal/LazyMapTestNative.kt to verify this is all working as expected.

saket:
Can try doing! Would adding a unit test that accesses `LazyMap` from a background thread be sufficient?

evant:
Haha, already done and merge 😉

saket:
haha nice!

PR #191, opened 2022-03-18
fix a few things in the new Compose docs

humblehacker:
- There were a few copy/paste errors where the `@Composable` was a subclass of `Fragment` rather than a function
- The modified version of the `viewModel()` function requires the template arg `VM` to be `reified`, otherwise you get the error
  > Cannot use VM as a reified type parameter. Use a class instead."
- I referenced the origin of the suggested `viewModel()` helper function, as this helped me understand what it was doing

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -222,7 +222,7 @@ class MyActivity : Activity() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


evant (approved):

evant:
Good catch!

PR #195, opened 2022-05-03
Update Kotlin, KSP, KotlinPoet, and Gradle

eygraber:


evant:
I should document this somewhere, but plan to be somewhat conservative on kotlin/ksp updates as the user can update them themselves. Not against updating, but it just needs to be for something we need instead of bumping arbitrarily.

Looks like  kotlinpoet: 1.10.2 didn't have a kotlin version bump so that should be safe to do.

eygraber:
I'd say that it might be worth it to keep up to date so that if there is a breaking change that requires a library update it'll be closer to that version. But there's no guarantee on that so ¯\_(ツ)_/¯

evant:
For the libs we are using those should be few and far between

eygraber:
@evant I'll be keeping this updated until it is time to merge :smile:

eygraber:
@evant I updated to Gradle 8.2.1 and Kotlin 1.9

This will probably help with bridging the gap towards Kotlin 2.0

eygraber:
Any idea why the linux build is failing? Does it need more memory?

ast/kapt/src/main/kotlin/me/tatarka/kotlin/ast/ModelAst.kt
```diff
@@ -660,12 +660,20 @@ private fun KmAnnotationArgument.asString(): String {
...
         is KmAnnotationArgument.ArrayValue -> "[${elements.joinToString { it.asString() }}]"
     }
 }
 
+private fun KmAnnotationArgument.ArrayKClassValue.asString(): String =
+    buildString {
+        repeat(arrayDimensionCount) { append("Array<") }
+        append(className)
+        repeat(arrayDimensionCount) { append(">") }
+    }
```
eygraber, line 675: Adopted from `ArrayKClassValue.toString`

.circleci/config.yml
```diff
@@ -51,14 +51,10 @@ commands:
 jobs:
   build-linux:
     docker:
-      - image: cimg/openjdk:11.0.11-browsers@sha256:02d8ac1beb99bff9c303376388355c6f87f9e29960804e920f01f30547214f7c
-    resource_class: medium+
+      - image: cimg/openjdk:17.0.7-browsers@sha256:52a950f68f31d7e947e9468e11acb38177272a59c446044d13f0c6d7d66d12ed
+    resource_class: large
```
eygraber, line 55: The daemon kept dying when running with medium+

I'm not sure if that would get fixed now since the cache is populated.

evant:
Fyi I was testing this locally and was running into issues with the kapt backend. After a bunch of debugging turns out it's running with the *embeded* gradle kotlin instead of the kotlin version you pull in. This means it was using kotlin 1.8.10 and kotlinx-metadata 0.5.0 which won't work with kotlin 1.9.0.

I think I'm just going to go ahead of get rid of the kapt backend, it's deprecated anyway. Was planning on doing a bugfix release before then but it's getting more and more of a pain to deal with ☹️.

build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
```diff
@@ -1,13 +1,19 @@
...
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 plugins {
     kotlin("jvm")
 }
 
 val libs = the<LibrariesForLibs>()
 
+kotlin {
+    jvmToolchain(17)
```
evant, line 11: I don't quite understand the distinction between the jvmToolchain and the target version. Is there a reason we need both and that they should be different?

build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
```diff
@@ -1,13 +1,19 @@
...
+}
+
+kotlin.compilerOptions.jvmTarget = JvmTarget.JVM_1_8
+
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
     targetCompatibility = JavaVersion.VERSION_1_8
```
evant, line 17: fyi cool with upping this to 11

build-logic/build.gradle.kts
```diff
@@ -1,15 +1,18 @@
+buildscript {
+    dependencies {
+        with(libs.kotlin.gradle.get()) {
+            classpath("$group:$name:$embeddedKotlinVersion")
+        }
+    }
+}
```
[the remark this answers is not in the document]
eygraber: Forcing Gradle to use Kotlin 1.9 results in errors, which I think has to do with https://kotlinlang.org/docs/whatsnew19.html#separate-compiler-plugins-for-official-kotlin-libraries but in any case Gradle has always said not to force a different version of Kotlin for their runtime because they offer no guarantees ([Gradle 8.3 should have Kotlin 1.9 support](https://github.com/gradle/gradle/pull/25794)).

build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
eygraber: The toolchain version specifies which version of Java should be used to compile the project. The target version specifies the least compatible version of bytecode that will be produced (i.e. projects using at least Java 8 can use this library).

build-logic/build.gradle.kts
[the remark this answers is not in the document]
evant: ah so this overrides the version of kotlin uses when compiling the build script logic right?

eygraber:
Are the kapt tests not run on CI? I didn't notice that they were failing locally.

I pushed a new commit addressing your comments.

build-logic/build.gradle.kts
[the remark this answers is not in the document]
eygraber: Correct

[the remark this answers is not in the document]
eygraber: https://docs.gradle.org/8.3-rc-1/release-notes.html

The rc releases sometimes have issues, but I've used them in the past with no problems. It should allow kapt to stay in if you really don't want to remove it yet.

If you do want to remove kapt, I have a stashed changeset that's most of the way there, and I can make a PR. Unless you want the cathartic experience of ripping it out 😅

evant:
>  I didn't notice that they were failing locally.

Hm, I wonder if it has something to do with the toolchain support. I'm running with java 20 so it's jumping to java 17. When I run with
```
JAVA_HOME=/usr/lib/jvm/java-17-openjdk ./gradlew test
```
it works fine.

Ok yeah it does have to do with toolchains. If I remove the toolchain lines and add back setting the source compatibility it works fine. I do wonder if the toolchian thing is worth it? In theory forcing a specific version of javac is good for consistency but running newer versions seems to work fine without it?

build-logic/build.gradle.kts
```diff
@@ -1,15 +1,18 @@
...
 dependencies {
-    implementation(libs.kotiln.gradle)
+    implementation(libs.kotlin.gradle)
     implementation(libs.detekt.gradle)
-    // hack to access version catelouge https://github.com/gradle/gradle/issues/15383
+    // hack to access version catalogue https://github.com/gradle/gradle/issues/15383
     implementation(files(libs.javaClass.superclass.protectionDomain.codeSource.location))
 }
-
-java {
```
evant, line 12: note: to run with java 20 I had to add
```
java {
    targetCompatibility = JavaVersion.VERSION_11
}

tasks.withType<KotlinCompile>().configureEach {
    compilerOptions.jvmTarget = JvmTarget.JVM_11
}
```
(17 probably would work as well)
but either way I think it's good to have for consistency.

eygraber:
TIL that toolchains aren't as good as I thought they were 😅

https://kotlinlang.slack.com/archives/C3PQML5NU/p1689790193010619?thread_ts=1689789822.225749&cid=C3PQML5NU

build-logic/build.gradle.kts
eygraber: Using a toolchain here would probably be a valid use case (and I tested that it works), but the alternative is to not use java 20 until Gradle supports it.

eygraber:
I pushed an update cleaning up some of the source and target usage, and removed the toolchains. Could possibly add the toolchain back conditionally in the future to aid in having a minimum required version of java.

The kapt failure when using java 20 with the toolchain set is odd; I'm going to file an issue with JetBrains.

evant (approved):

build-logic/build.gradle.kts
evant: > but the alternative is to not use java 20 until Gradle supports it

yeah true, I might end up adding a target here because it just means you'd be generating older class files than needed so it's not a huge issue, but it's not a priority or anything.

PR #196, opened 2022-05-06
Fix for Redeclaration errors due to KSP source set code generation

### LINKED ISSUE #194: Redeclaration errors for `@Component` when ksp is used for common and target source sets
I have a component written in `commonMain`:

```
@Component
abstract class AppComponent(
  @Component protected val platformComponent: PlatformComponent
) {
  companion object
}
```

I also have another component written in `commonMain` as an `expect` class:

```
expect abstract class PlatformComponent {
  companion object
}
```
and an `actual` class in my targets (e.g. android):

```
@Component
actual abstract class PlatformComponent(
  @get:Provides val context: AppContext
) {
  actual companion object
}
```

I set up ksp to run for common and each of my targets:

```
kotlin.sourceSets.commonMain {
  kotlin.srcDir("build/generated/ksp/metadata/commonMain/kotlin")
}

kotlin.targets.configureEach {
  kotlin.configureKsp(name)
}

fun KotlinSourceSetContainer.configureKsp(targetName: String) {
  runCatching {
    dependencies {
      add("ksp${targetName.capitalized()}", libs.kotlinInject.compiler)
    }

    sourceSets.configureEach {
      kotlin.srcDir("build/generated/ksp/$targetName/$name/kotlin")
    }
  }
}

dependencies {
  add("kspCommonMainMetadata", libs.kotlinInject.compiler)

  tasks.withType<KotlinCompile>().configureEach {
    if(name != "kspCommonMainKotlinMetadata") {
      dependsOn("kspCommonMainKotlinMetadata")
    }
  }
}
```

And I get the following errors:

```
e: /home/eli/workspace/galgal/core/build/generated/ksp/android/androidDebug/kotlin/com/eygraber/galgal/core/components/InjectAppComponent.kt: (9, 14): Redeclaration: InjectAppComponent

e: /home/eli/workspace/galgal/core/build/generated/ksp/metadata/commonMain/kotlin/com/eygraber/galgal/core/components/InjectAppComponent.kt: (6, 14): Redeclaration: InjectAppComponent
```

eygraber:
Fixes #194

Stacked on #195

### DIFF
--- a/ast/ksp/build.gradle.kts
+++ b/ast/ksp/build.gradle.kts
@@ -13,5 +13,5 @@ dependencies {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    kotlinOptions.freeCompilerArgs += "-Xopt-in=com.squareup.kotlinpoet.ksp.KotlinPoetKspPreview"
+    kotlinOptions.freeCompilerArgs += "-opt-in=com.squareup.kotlinpoet.ksp.KotlinPoetKspPreview"
 }
--- a/build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
@@ -1,6 +1,4 @@
-import org.jetbrains.kotlin.gradle.dsl.KotlinJsCompile
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithTests
-import org.jetbrains.kotlin.gradle.targets.js.testing.KotlinJsTest
 import org.jetbrains.kotlin.gradle.targets.js.npm.tasks.KotlinNpmInstallTask
 
 plugins {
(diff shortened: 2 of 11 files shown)


evant:
Honestly this seems too hacky for me to accept, I think it would be better to wait for ksp to support multiplatform projects in a nicer way, I am a fan of the suggestion [here](https://github.com/google/ksp/issues/567#issuecomment-967711207). In the meantime, can't you exclude running ksp on `kspCommonMainMetadata`? I feel like that would get you closer to what you want.

eygraber:
Yeah there's a lot of hackiness + I just realized it would break for hierarchical layouts.

Excluding `kspCommonMainMetadata` works; I was probably a little thrown off by that initially because the generated source sets can't see the common source set in the IDE, but everything builds fine. That's definitely a better solution for now, thanks!

eygraber:
@evant excluding `kspCommonMainMetadata` works until I need to process classes from both common and targets, and reference the generated common code from common. 

I could expect/actual everything, but that would be very inconvenient for my use case. Any other ideas?

PR #197, opened 2022-05-15
Ignore local properties

eygraber:


evant:
This feels kinda arbitrary since this isn't an android project. You can always add extra things to your [global gitignore](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) if it's bothering you.

eygraber:
True, but I usually use Android Studio for non Android projects, if I already have it open for an Android project. 

I'm not a big fan of global gitignores because of previous bad experiences with lost changes 😅

PR #199, opened 2022-05-17
Leak detector

### LINKED ISSUE #188: Leak when child dependency is provided to a parent dependency that has a longer lifecycle
If a child dependency is provided to a parent dependency, and that parent has a longer lifecycle than the child (e.g. the parent is scoped to an Android application, and the child is scoped to an Android activity), then the child dependency can leak (in the example below, the activity context is leaked):

```
@Inject
class ResolverWrapper(val resolver: ContentResolver)

@Singleton
@Component abstract class ParentComponent {
  @Provides @Singleton fun providesContentResolver(
    context: Context
  ): ContentResolver = context.contentResolver
}

@Component abstract class ChildComponent(
  @Component val parent: ParentComponent,
  @get:Provides val activityContext: Context
) {
  abstract val wrapper: ResolverWrapper
}
```

eygraber:
Fixes #188 

@evant what do you think of this general approach? My thinking is that if we're injecting into something that's scoped, and the thing we're injecting is not in that scope, we consider it a leak. It could be opt in/out, and there could also be a new annotation (e.g. `AllowScopeLeak`) for cases where there is nothing wrong with leaking it.

Obviously still a lot to do here, but I wanted to get your opinion before going too far.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,6 @@
 /build
 /.idea
 *.iml
+
+# For Android Studio
+local.properties
--- a/ast/ksp/build.gradle.kts
+++ b/ast/ksp/build.gradle.kts
@@ -13,5 +13,5 @@ dependencies {
 }
 
 tasks.withType<KotlinCompile>().configureEach {
-    kotlinOptions.freeCompilerArgs += "-Xopt-in=com.squareup.kotlinpoet.ksp.KotlinPoetKspPreview"
+    kotlinOptions.freeCompilerArgs += "-opt-in=com.squareup.kotlinpoet.ksp.KotlinPoetKspPreview"
 }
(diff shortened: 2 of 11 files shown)


evant:
Yeah I've given this a bit of thought. As long as you are looking 'up' the graph you are fine, as any parent components are expected to live longer anyway. The problem is when you look 'down' the graph. We could disallow this completely but I think it would be ok to allow it with unscoped dependencies? Since those don't imply any particular lifecycle.

In either case, the problem right now is that the information on where in the graph a dependency comes from is dropped when collecting into a map so that needs to be re-thought.

eygraber:
> Since those don't imply any particular lifecycle

I think that's why it's fine allowing them to go down the graph, whether they're scoped or not.

Do you think leaking dependencies collecting into a map can be handled separately in a later PR, or should it all go out together?

evant:
I put up a [pr](https://github.com/evant/kotlin-inject/pull/201) with a fix. If you are injected a dependency with a scope into another dependency with a scope it'll ensure the dependency's scope lifetime is at least at long.

eygraber:
Awesome, just happy to have the functionality 😁

PR #216, opened 2022-07-29
Removed ignoreFailures = true to not miss failing tests

evant:


evant:
Tests where already failing but where covered up, merging in to represent the actual state of things.

PR #217, opened 2022-08-02
Use a NameAllocator to prevent arg name shadowing

PaulWoitaschek:
We make use of allWarningsAsErrors. This has lead to errors from the generated code because the (arg) names are shadowed.

This PR ensures that there are unique argument names and also adds a test case (that fails without the new logic).

Additionally, this PR enables allWarningsAsErrors for the ksp module in order to prevent future regressions.

As there was no code style attached to this project and I did not want to manually format the code by hand, I just let IntelliJ auto formatting the files that I touched.

### Previously
```kotlin
package me.tatarka.inject.test

import kotlin.reflect.KClass

public fun KClass<NestedNestedFunctionComponent>.create(): NestedNestedFunctionComponent =
    InjectNestedNestedFunctionComponent()

public class InjectNestedNestedFunctionComponent() : NestedNestedFunctionComponent() {
  public override val personFactory: Person.Factory
    get() = Person.Factory(
      create = { arg0 ->
        Person(
          house = House.Factory(
            create = { arg0 ->
              House(
                bricks = arg0
              )
            }
          ),
          name = arg0
        )
      }
    )
}
```

### New
```kotlin
package me.tatarka.inject.test

import kotlin.reflect.KClass

public fun KClass<NestedNestedFunctionComponent>.create(): NestedNestedFunctionComponent =
    InjectNestedNestedFunctionComponent()

public class InjectNestedNestedFunctionComponent() : NestedNestedFunctionComponent() {
  public override val personFactory: Person.Factory
    get() = Person.Factory(
      create = { arg0 ->
        Person(
          house = House.Factory(
            create = { arg0_ ->
              House(
                bricks = arg0_
              )
            }
          ),
          name = arg0
        )
      }
    )
}
```


integration-tests/common/src/test/kotlin/me/tatarka/inject/test/FunctionTest.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("unused")
```
evant, line 1: I'd be fine with getting rid of unused variables in tests, but it's a pretty minor thing.

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/FunctionTest.kt
```diff
@@ -8,72 +10,119 @@ import me.tatarka.inject.annotations.Component
...
+@Component
+abstract class NestedNestedFunctionComponent {
+    abstract val personFactory: Person.Factory
 }
 
 class FunctionTest {
 
-    @Test fun generates_a_component_that_provides_a_function_that_returns_a_dep() {
+    @Test
+    fun nestedNamesGenerated() {
```
evant, line 88: minor: could you update the test name to follow the pattern of other tests?

evant (commented):
Thanks for this! I knew it was an issue but hadn't actually run into it in practice yet so was low priority to fix. Have some minor comments but overall looks good!

> As there was no code style attached to this project and I did not want to manually format the code by hand, I just let IntelliJ auto formatting the files that I touched.

Yep that's a good strategy, I should probably note somewhere that we want to be using the default IntellijJ format settings. I was a fan of keeping annotations on the same line but it's an uphill-battle fighting the defaults so eh. Note: we do have detekt enabled which does catch some issues but it's a bit lax.

evant:
There's some merge conflicts that need to be fixed, let me know if you have any questions about them.

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/FunctionTest.kt
PaulWoitaschek: Yep, done 👍

integration-tests/common/src/test/kotlin/me/tatarka/inject/test/FunctionTest.kt
PaulWoitaschek: I replaced the file supressions with individual ones.
The thing is that they are not actually unused but necessary for the test setup.

PaulWoitaschek:
> There's some merge conflicts that need to be fixed, let me know if you have any questions about them.

Was pretty straight forward; merged main.

> Thanks for this! I knew it was an issue but hadn't actually run into it in practice yet so was low priority to fix. Have some minor comments but overall looks good!

Yeah we just did and had to setup a large dependency chain by hand 😃

Issue #164, opened 2021-11-06
Proposal: make assisted arguments explicit


saket:
kotlin-inject's way of injecting dependencies with assisted arguments is really clever but I've been thinking it's very implicit and hidden for readers. Thoughts on making it explicit in some manner? Using dagger-like `@Assisted` annotation is one option, but I'm not sure if you'd want that. It'll also remove the restriction of keeping assisted arguments at the end of a constructor, which is another hidden restriction.

evant:
So one thought I had around this is to recommend typealiases when you are doing assisted injection.

```kotlin
typealias FooFactory = (Baz) -> Foo
@Inject class Foo(injectedArg: Bar, assistedArg: Baz)
...
@Inject class UsesFoo(fooFactory: FooFactory)
```

Since the typealias is in the same file, you can see which args are assisted as they are the ones explicitly declared in the typealias. It also helps with refactoring if you ever need to change the signature.

> It'll also remove the restriction of keeping assisted arguments at the end of a constructor, which is another hidden restriction.

I will say, I ended up quite liking this restriction, it makes the argument ordering consistent. Curious if you had any examples where this ordering wouldn't be ideal?

saket:
> I will say, I ended up quite liking this restriction, it makes the argument ordering consistent. Curious if you had any examples where this ordering wouldn't be ideal?

I'd say this is a really subjective opinion. I know that the preference of ordering varies by person. For example, some people like keeping annotated values towards the end for visual hierarchy.

PaulWoitaschek:
We always use named arguments for almost everything so that changing the position doesn't break existing code.
When developers add a dependency to a class, they usually add it at the last position. If now that dependency was managed by a kotlin-inject factory, this will break the DI setup.

I've seen a few times now, that developers got really confused: "Hey, can you help me? I added that class as a dependency and kotlin-inject says that it can't create the class, but that dependency should be definitely available".

I would really prefer an explicit approach here that makes it clear which arguments of a class are assisted and which should be available in the DI graph.

evant:
One of the benefits of the current approach is the client gets to decide which dependencies to inject vs providing manually. I'm not sure that actually ever comes up in practice though so may not be worth it over explicit documentation.

PaulWoitaschek:
That case could still be covered with a secondary constructor if that really ever comes up, right?

evant:
Ok here's my proposal:

# Introduce an `@Assisted` annotation to make assisted arguments explicit

[...]

PR #227, opened 2022-08-16
Initial pass at requiring @Assisted on assisted parameters

evant:
#164


PaulWoitaschek:
A test case where assistet params are not in the end would be great to have. 
Also, a test case that validates the exception message would be great when there are missing bindings.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
```diff
@@ -61,30 +61,105 @@ class TypeResultResolver(private val provider: AstProvider, private val options:
...
                 throw FailedToGenerateException(cannotFind(key))
             }
         }
+
+        if (resolvedImplicitly.isNotEmpty()) {
+            provider.warn(
+                """
+                Implicit assisted parameters are deprecated and will be removed in a future version.
+                Annotate the following with @Assisted: [${
+                resolvedImplicitly.joinToString { (name, type) -> "$name: ${type.simpleName}" }
```
PaulWoitaschek, line 156: Is it possible to reference line numbers? I had the experience a few times where a clickable line number would have been helpful to jump directly to the problematic code.

evant: By passing the element to the logger we should be getting clickable line numbers in the error message. If this is not working in some cases it would be worth looking into.

evant:
@PaulWoitaschek agreed on the test cases. In fact the primary reason this is just in draft is because it needs more tests!

PaulWoitaschek (approved):

PR #231, opened 2022-08-29
Make arg names not unique accross a component

PaulWoitaschek:
This is a follow up of https://github.com/evant/kotlin-inject/pull/217

Create an ArgNameAllocator and reset it between provider calls. Else, the arg names would be unique across all functions of a component.

### DIFF
--- a/kotlin-inject-compiler/core/src/test/kotlin/me/tatarka/inject/compiler/ArgNameAllocatorTest.kt
+++ b/kotlin-inject-compiler/core/src/test/kotlin/me/tatarka/inject/compiler/ArgNameAllocatorTest.kt
@@ -9,26 +9,22 @@ internal class ArgNameAllocatorTest {
     @Test
     fun `same indices lead to different arg names`() {
         val allocator = ArgNameAllocator()
-        allocator.assert(1, "arg1")
-        allocator.assert(1, "arg1_")
+        assertThat(allocator.newName(1)).isEqualTo("arg1")
+        assertThat(allocator.newName(1)).isEqualTo("arg1_")
     }
 
     @Test
     fun `reset leads to same arg names`() {
         val allocator = ArgNameAllocator()
-        allocator.assert(1, "arg1")
+        assertThat(allocator.newName(1)).isEqualTo("arg1")
         allocator.reset()
-        allocator.assert(1, "arg1")
+        assertThat(allocator.newName(1)).isEqualTo("arg1")
     }
 
     @Test
     fun `different indices lead to different arg names`() {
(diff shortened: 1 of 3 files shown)


kotlin-inject-compiler/core/src/test/kotlin/me/tatarka/inject/compiler/ArgNameAllocatorTest.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Test
+    fun `different indices lead to different arg names`() {
+        val allocator = ArgNameAllocator()
+        allocator.assert(0, "arg0")
+        allocator.assert(1, "arg1")
+    }
+
+    private fun ArgNameAllocator.assert(index: Int, expected: String) {
+        assertThat(newName(index), "arg$expected")
```
evant, line 31: The way to use [assertk](https://github.com/willowtreeapps/assertk) is to call methods on an `assertThat` call. The call itself won't actually assert anything. For example:
```kotlin
assertThat(allocator.newName(0)).isEqualTo("arg0")
```

Also, there are ways to abstract out [common assertions](https://github.com/willowtreeapps/assertk#custom-assertions) but honestly for this usecase I don't think it's worth it.

evant (changes requested):

kotlin-inject-compiler/core/src/test/kotlin/me/tatarka/inject/compiler/ArgNameAllocatorTest.kt
evant: Sorry should've been more clear on my second point. I don't think the extension method is worth it and would rather see direct calls to `assertThat(...).isEqualTo(...)`, I think that's more clear imo.

PaulWoitaschek: I thought you thout think that custom assertions are not worth it.

To me that's duplicated code, but your call - pushed.

evant (approved):

kotlin-inject-compiler/core/src/test/kotlin/me/tatarka/inject/compiler/ArgNameAllocatorTest.kt
evant: Wouldn't mind discussing this more, but the basic idea is clarity can be more important de-duplication particularly in tests. This example is pretty minor, but you still have to look at the implementation of the extension to see what exactly it's testing and in doesn't really save that much.

PR #232, opened 2022-09-06
Add a test case for overwriting provides functions

PaulWoitaschek:
This test case is currently failing. It currently leads to a stack overflow at runtime due to this code (example) being generated:
```kotlin
public class InjectInterfaceComponentWithIdenticalProvides(
  foo: Foo
) : InterfaceComponentWithIdenticalProvides(foo) {
  public override val foo: Foo
    get() = foo
}
```

The code that should be generated instead would be:

```kotlin
public class InjectInterfaceComponentWithIdenticalProvides(
  foo: Foo
) : InterfaceComponentWithIdenticalProvides(foo)
```


evant:
Good catch! built off of this with https://github.com/evant/kotlin-inject/pull/233 to fix the issue.

PR #234, opened 2022-09-07
Generate graph when @Inject is typealiased

jstarczewski:
Hey @evant 
I have a use case where I want to use `kotlin-inject` in the project, but do not really want to spread the `@Inject` annotation all over the project, so I thought about using a `typealias` for `@Inject`, but I encountered some errors when using KSP (works with KAPT). Simple example below:
```Kotlin
import com.softwaret.booksapp.books.service.BooksApiService

typealias InjectTypealias = Inject

@InjectTypealias
internal class GetBooksUseCase(
    private val booksApiService: BooksApiService
)
```
Error (part of it)
```
[ksp] Cannot find an @Inject constructor or provider for: com.softwaret.booksapp.books.usecase.GetBooksUseCase
```
I do not know whether it is intended behaviour, but I tried to fix it myself. Please, review the code and the tests. I also do not know whether I put the tests in the correct place, or if it is enough (for a fix and tests).

The change I did seems to work for `@Inject` but does not work for type-aliasing `@Component` annotations (and maybe other ones), but I did no further investigation of the process because of lack of time and actually type aliasing `@Component` and other annotations is not my use case, but I am interested in doing further investigations and maybe more general fix for that if you think it is needed.


kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
```diff
@@ -263,4 +263,57 @@ class RoundsTest {
             doesNotContain("Cannot provide", "as it is already provided")
         }
     }
+
+    @Test
+    fun can_generate_for_inject_typealias_on_primary_constructor() {
```
evant, line 268: This class is for running the process in multiple rounds. You should add an example to the [integration tests](https://github.com/evant/kotlin-inject/tree/main/integration-tests/common/src/test/kotlin/me/tatarka/inject/test) instead. You'll find the majority of tests are in there.

evant (changes requested):
The reason it doesn't work for `@Component` is this visitor that searches for it would have to be updated as well. I had originally omitted resolving type aliases as it's a slight extra performance cost but I do think it's a reasonable thing to do.

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
jstarczewski: I moved the tests to separate file in integration tests module, and added two more tests. I will try to make type-aliasing `@Component` work in a separate pull requests, so this one will not grow more.

ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSUtil.kt
```diff
@@ -26,10 +27,17 @@ internal fun KSAnnotated.hasAnnotation(packageName: String, simpleName: String):
...
+        shortName.asString() == simpleName -> declaration.packageName.asString() == packageName
+        declaration is KSTypeAlias -> declaration.isTypeAliasForName(packageName, simpleName)
+        else -> false
+    }
+}
+
+private fun KSTypeAlias.isTypeAliasForName(packageName: String, simpleName: String): Boolean {
+    val aliasedType = type.resolve()
+    return aliasedType.declaration.packageName.asString() == packageName && aliasedType.toString() == simpleName
 }
```
evant, line 41: You can have a typealis point to a typealias so this should probably recurse. I think you could also consolidate the 2 name checks by doing that.

```suggestion
    return when {
        declaration is KSTypeAlias -> declaration.type.resolve().declaration.hasName(packageName, simpleName)
        else ->  shortName.asString() == simpleName && declaration.packageName.asString() == packageName
    }
```

jstarczewski: Yes, you are right about the type alias pointing to type alias case. Fixed and added the test for it.  For now, I refactored the `isTypeAliasForName` function to start recursion with `tailerc` modifier to share the stack frame. I also changed the order of name checks in this function, to check the name of the aliased type first.

jstarczewski: And I do not know whether I get you right, but calling `hasName` like suggested here `declaration.type.resolve().declaration.hasName(packageName, simpleName)` on `declaration` after `resolve()` is not possible because we do not know if this `declaration` is a `KSAnnotation`, so I left the check for type alias in a separate function instead of trying to refactor for some generic solution, because it looked right and readable to me.

evant: Sorry yeah gave that suggestion without actually checking it was correct, this should work though:

```kotlin
private fun KSAnnotation.hasName(packageName: String, simpleName: String): Boolean {
    return annotationType.resolve().declaration.hasName(packageName, simpleName)
}

private tailrec fun KSDeclaration.hasName(packageName: String, simpleName: String): Boolean =
    when (this) {
        is KSTypeAlias -> type.resolve().declaration.hasName(packageName, simpleName)
        else -> this.simpleName.asString() == simpleName && this.packageName.asString() == packageName
    }
```

jstarczewski: @evant Now it works, changed as you suggested, thanks. Just to be sure, I see that there is also `Util.kt` file with similar implementation of `KSAnnotation.hasName` triggered from `KSAnnotated.hasAnnotation`, but as I check the implementation and usage, it is related to `@Provides` and `@Component` annotations, which I plan to tackle later on in separate PR, so I think we can let it be the way it is for now?

evant: Yeah that sounds right

evant (approved):

PR #257, opened 2023-04-09
fix some typos

zsqw123:


README.md
```diff
@@ -224,7 +224,7 @@ fun myFunction(dep: Dep) {
 }
 ```
 
-You can then use the type alias anywhere and you will be provided with a function that calls the top-level one with the
+You can then use the type alias anywhere, and you will be provided with a function that calls the top-level one with the
```
evant, line 227: I'm not sure the comma here is necessary?

README.md
```diff
@@ -248,7 +248,7 @@ fun myFunction(dep: Dep, arg: String): String = ...
 
 ### Scopes
 
-By default kotlin-inject will create a new instance of a dependency each place it's injected. If you want to re-use an
+By default, kotlin-inject will create a new instance of a dependency each place it's injected. If you want to re-use an
```
evant, line 251: Same here

evant (commented):
Good catches!

README.md
zsqw123: ok, maybe I can revert it, I will only fix some typos :)

evant (approved):

PR #275, opened 2023-07-12
Update assertk

eygraber:


kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/Utils.kt
```diff
@@ -11,4 +14,10 @@ import javax.tools.Diagnostic
...
 fun Assert<TestCompilationResult>.warnings(): Assert<String> =
-    prop("warnings") { it.output(Diagnostic.Kind.WARNING) }
\ No newline at end of file
+    prop("warnings") { it.output(Diagnostic.Kind.WARNING) }
+
+internal inline fun <T> assertSuccess(block: () -> T) =
+    assertThat(runCatching(block)).isSuccess()
+
+internal inline fun <T> assertFailure(block: () -> T) =
+    assertThat(runCatching(block)).isFailure()
```
evant, line 23: You shouldn't need to add wrappers for these. `assertFailue` should exist in assertk and
```kotlin
assertThat { ... }.isSuccess()
```
can be replaced with just
```kotlin
assertThat(...)
```

eygraber: I think that was deprecated

eygraber: Oh nevermind, I see what you're saying

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
```diff
@@ -41,13 +40,13 @@ class RoundsTest {
                     }
                 """.trimIndent()
             ).compile()
-        }.isSuccess()
+        )
```
evant, line 43: another assertThat without an assertion

evant: oh nvm it's the same one

evant (approved):

PR #276, opened 2023-07-12
Address warnings

eygraber:


build-logic/src/main/kotlin/kotlin-inject.publish.gradle.kts
```diff
@@ -51,6 +51,7 @@ publishing {
         val publishApple by tasks.registering {
             publications.all {
                 if (name.contains(Regex("macos|ios|tvos|watchos"))) {
+                    @Suppress("DEPRECATION")
                     dependsOn("publish${name.capitalize(Locale.ROOT)}PublicationToSonatypeRepository")
```
evant, line 55: This can be replaced with `name.replaceFirstChar { it.uppercase() }`

build-logic/settings.gradle.kts
```diff
@@ -1,6 +1,6 @@
 dependencyResolutionManagement {
     versionCatalogs {
-        val libs by creating {
```
evant, line 3: is this syntax not recommended or something?

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -241,6 +241,7 @@ fun AstMember.isProvider(): Boolean =
 fun AstClass.toInjectName(): String =
     "Inject${toClassName().simpleNames.joinToString("_")}"
 
+@Suppress("DEPRECATION")
 fun AstType.toVariableName(): String =
     simpleName.split(".")
         .joinToString("_") { it.decapitalize(Locale.US) } +
```
evant, line 247: can be replaced by `{ name -> name.replaceFirstChar { it.lowercase() } }`

build-logic/src/main/kotlin/kotlin-inject.publish.gradle.kts
eygraber: True, I'm still a little salty that `capitalize` was deprecated. I'll make the change

build-logic/settings.gradle.kts
eygraber: It leads to warnings about not being used which pollute the output. I think this syntax is fine if the property is actually used.

evant: heh, looks like people have been complaining about this for a while https://youtrack.jetbrains.com/issue/KT-38871/Kotlin-Gradle-DSL-MPP-UNUSEDVARIABLE-when-configuring-a-sourceset-with-delegated-property. I'd rather use a consistent syntax then change it up to just avoid a warning. What about just putting a `@file:Suppress("UNUSED_VARIABLE")` at the top of the file?

eygraber: It actually looks like this was [addressed in Kotlin 1.9](https://github.com/jetbrains/kotlin/commit/681e85eaed8e047ea955cbc8b3ea5037b1581cf9) so Gradle 8.3 should fix this. 

I'll test with the rc, and if that actually fixes it, I'll suppress these and make a follow up PR to remove them once 8.3 is released.

build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
```diff
@@ -34,8 +34,9 @@ kotlin {
                 dependsOn(nativeTest)
             }
         }
+
         // Ensure xml test reports are generated
-        val jvmTest by tasks.getting(Test::class) {
+        tasks.named("jvmTest", Test::class).configure {
```
eygraber, line 39: This case is also not just about the unused variable warning, but because `getting` doesn't use task configuration avoidance. I'll leave my changes for this, unless you feel I should revert it.

evant: Oh dang I thought it did since it was a 'newer' api.

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/WarningTest.kt
```diff
@@ -35,6 +35,7 @@ class WarningTest {
 
     @ParameterizedTest
     @EnumSource(Target::class)
+    @Suppress("UNUSED_PARAMETER")
     fun warns_on_implicit_assisted_params(target: Target) {
         val projectCompiler = ProjectCompiler(Target.KAPT, workingDir)
```
eygraber, line 40: Does this test need to be parameterized? It looks like `KAPT` is hardcoded in.

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
```diff
@@ -41,13 +40,13 @@ class RoundsTest {
                     }
                 """.trimIndent()
             ).compile()
-        }.isSuccess()
+        )
```
evant, line 43: since we aren't actually asserting anything `assertThat` isn't doing anything here. Maybe add a comment that we expect it not to throw instead

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/WarningTest.kt
evant: good catch! It might have been a copy/paste error. I'd say try it and if it's failing for ksp for some reason we can revisit that in a different pr.

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
eygraber: How about `assertDoesNotThrow`?

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/WarningTest.kt
eygraber: Works with ksp :tada:

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/RoundsTest.kt
evant: I mean if it throws it'll fail the test anyway, no strong opinions though

eygraber: OK pushed update with it. I think `assertDoesNotThrow` provides a better UX vs just throwing.

evant (approved):

eygraber:
Should be ready

PR #277, opened 2023-07-12
Add support for WASM

eygraber:
Blocked by https://github.com/willowtreeapps/assertk/pull/493


kotlin-inject-runtime/src/jsAndWasmJsMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
eygraber, line 1: Made this a shared source set for `js` and `wasmJs`. The IDE shows some errors, but it seems to be building fine.

evant:
> Blocked by https://github.com/Kotlin/kotlinx.coroutines/pull/3849

fyi, since this only needed for tests, it would be acceptable to skip those tests for the wasm target while it's still experimental

eygraber:
I don't think it's possible to skip coroutine tests for wasm because atomicfu and coroutines artifacts are clashing. Once the new version of coroutines releases, the commit commenting out coroutines can be reverted and this should be able to be merged.

kotlin-js-store/yarn.lock
eygraber, line 1: I added this because I kept having to make sure that I wasn't committing it. I think best practice is to track it in git, but I'll remove if you don't want it.

evant: > I think best practice is to track it in git

If that's the case, then yeah it's fine to keep, I wasn't sure what was best practice here

evant:
So, in terms of skipping for wasm my thought would be to move the tests that use coroutines to a source set that excludes wasm, and the wasm target would be the one that doesn't include the coroutine dependency. It shouldn't be bringing in the conflicting dependencies at the same time right? Is there some reason that wouldn't work?

eygraber:
> Is there some reason that wouldn't work?

It should work, but it might be worth waiting for the coroutines release. I'll check in with the kotlinx team to see if they have any kind of timeline.

kotlin-js-store/yarn.lock
evant: https://kotlinlang.org/docs/js-project-setup.html#version-locking-via-kotlin-js-store

:+1: found it

evant:
🤔 https://github.com/Kotlin/kotlinx.coroutines/releases/tag/1.8.0-RC

eygraber:
Yup, going to try updating to that soon. Are you ok with releasing the RC?

eygraber:
OK should be ready to go. How do you feel about merging with coroutine 1.8-RC?

evant:
It's not something we are distributing with so it should be fine

evant (approved):

PR #280, opened 2023-07-20
Use Gradle's lazy APIs when possible

eygraber:


build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
```diff
@@ -19,26 +20,12 @@ kotlin {
...
-        for (sourceSet in nativeTargets) {
-            getByName("${sourceSet}Main") {
-                dependsOn(nativeMain)
-            }
-            getByName("${sourceSet}Test") {
-                dependsOn(nativeTest)
-            }
-        }
+    @OptIn(ExperimentalKotlinGradlePluginApi::class)
+    targetHierarchy.default()
```
eygraber, line 24: This is a really cool feature that sets up the source sets the way you'd expect them to be (e.g. `native` is set up), and it is extensible.

It is experimental, so I can remove it if you want.

evant: kotlin multiplatform is experimental anyway so it's fine. Maybe try a quick `./gradlew publichToMavenLocal` to make sure there isn't any issues with publishing with the new layout.

evant (commented):
registering, existing, gradle is running out of words aren't they? lol

evant:
hey just an fyi, when you stack pr's it's nicer to have the stacked pr point to the original as the base branch instead of main, it makes the diff easier to read. github will automatically retarget when the original one gets merged.

eygraber:
I think the issue with not pointing at main is that it keeps the PR in my fork, and I don't think it's possible to re-point it to this repo after the target gets merged.

build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
eygraber: LGTM although I'm building on Linux so only the linuxX64 target gets built for native.

evant:
bah yeah you are right

build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
evant: seems good on macos too

evant (approved):

PR #323, opened 2023-11-23
Update Kotlin to 1.9.21

eygraber:


build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
```diff
@@ -14,15 +13,22 @@ kotlin {
...
+    iosArm64()
+    iosX64()
+    iosSimulatorArm64()
+    tvosArm64()
+    tvosX64()
+    tvosSimulatorArm64()
+    watchosArm32()
+    watchosArm64()
+    watchosX64()
+    watchosSimulatorArm64()
```
eygraber, line 28: preset APIs were deprecated.

build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
```diff
@@ -14,15 +13,22 @@ kotlin {
...
 
     jvm()
 
-    @OptIn(ExperimentalKotlinGradlePluginApi::class)
-    targetHierarchy.default()
```
eygraber, line 24: This is now the default and doesn't need to be explicitly called

kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
```diff
@@ -1,10 +1,8 @@
 package me.tatarka.inject.internal
 
-import kotlin.native.concurrent.SharedImmutable
 import kotlinx.atomicfu.locks.ReentrantLock
 import kotlinx.atomicfu.locks.withLock
 
-@SharedImmutable
```
eygraber, line 7: Deprecated, message says that it doesn't actually do anything

evant:
> Also might be worth not merging until KSP has support for Kotlin 1.9.21 which should come concurrently with the fix for https://github.com/google/ksp/issues/1591

Yeah looks like the tests are failing because of this so think we'll have to wait

evant:
https://github.com/google/ksp/releases/tag/1.9.21-1.0.15

eygraber:
Updated

evant (approved):

PR #326, opened 2023-11-27
better handle duplicate member declarations in superclasses

### LINKED ISSUE #309: Conflicting declarations with multiple identical overrides
Code such as the following:

```kotlin
class MyDep

interface Provisioner {

    @Provides
    fun provideMyDep() = MyDep()
}

interface Accessor1 {

    val myDep: MyDep
}

interface Accessor2 {

    val myDep: MyDep
}

@Component
interface MyComponent : Provisioner, Accessor1, Accessor2
```

Generates the following:

```kotlin
public fun KClass<MyComponent>.create(): MyComponent = InjectMyComponent()

public class InjectMyComponent : MyComponent {
  public override val myDep: MyDep
    get() = provideMyDep()

  public override val myDep: MyDep
    get() = provideMyDep()
}
```

The override should only be generated once, as this code cannot be compiled:

```
Conflicting declarations: public open val myDep: MyDep, public open val myDep: MyDep
```

evant:
Fixes #309

### DIFF
--- a/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
+++ b/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
@@ -101,6 +101,8 @@ sealed class AstMember : AstElement(), AstAnnotated, AstHasModifiers {
 
     abstract fun overrides(other: AstMember): Boolean
 
+    abstract fun signatureEquals(other: AstMember): Boolean
+
     abstract val returnType: AstType
 
     abstract fun returnTypeFor(enclosingClass: AstClass): AstType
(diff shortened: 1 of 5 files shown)


ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
```diff
@@ -133,8 +135,19 @@ abstract class AstFunction : AstMember() {
 
     abstract override fun hashCode(): Int
 
-    override fun toString(): String {
-        return "$name(${parameters.joinToString(", ")}): $returnType"
+    override fun toString(): String = buildString {
```
evant, line 138: make sure we include the receiver for these.

PR #339, opened 2024-01-26
include multiplatform specific information

RivuChk:
I felt set up information was missing for multiplatform projects, I obtained these by viewing multiple samples and from kotlin-slack, so decided to add them.
Please do let me know if I missed something or if something is not correct.

### DIFF
--- a/docs/multiplatform.md
+++ b/docs/multiplatform.md
@@ -1,0 +1,89 @@
+# Multiplatform
+
+This document covers using kotlin-inject in a multiplatform project.
+
+## Download
+
+Using [ksp](https://github.com/google/ksp)
+
+`settings.gradle`
+
+```kotlin
+pluginManagement {
+    repositories {
+        gradlePluginPortal()
+        mavenCentral()
+    }
+}
+
+dependencyResolutionManagement {
+    repositories {
+        google()
+        mavenCentral()
+    }
+}
+```
+
+`build.gradle`
+
+Add Plugin
+```kotlin
+plugins {
+    id("org.jetbrains.kotlin.multiplatform") version "1.9.21"
+    id("com.android.library")
+    id("com.google.devtools.ksp") version "1.9.21-1.0.16"
+}
+```
+
+Add Runtime dependency in commonMain
+```kotlin
+sourceSets {
+    androidMain.dependencies {
+        //...
+    }
+    commonMain.dependencies {
+        implementation("me.tatarka.inject:kotlin-inject-runtime:0.6.3")
+    }
+}
+```
+If you're targetting iOS, you need to add each of the chipset as separate sourcesets to include generated files.
+Same goes to any other LLVM (MacOS, TVOS etc.) you're targetting.
+```kotlin
+val iosX64Main by getting {
+    dependencies {
+        dependsOn(iosMain.get())
+        kotlin.srcDir("build/generated/ksp/iosX64/iosX64Main/kotlin")
+    }
+}
+val iosArm64Main by getting {
+    dependencies {
+        dependsOn(iosMain.get())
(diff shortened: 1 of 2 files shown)


RivuChk:
Hey @evant 
Did you have a chance to review it? Is there anything that I should change/add in terms of Multiplatform documentation?
Please do let me know, sorry if I bothered you.

docs/multiplatform.md
```diff
@@ -0,0 +1,89 @@
+# Multiplatform
+
+This document covers using kotlin-inject in a multiplatform project.
```
evant, line 3: looks like this is specifically for android/iOS, can we call that out?

docs/multiplatform.md
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+```
+If you're targetting iOS, you need to add each of the chipset as separate sourcesets to include generated files.
+Same goes to any other LLVM (MacOS, TVOS etc.) you're targetting.
+```kotlin
+val iosX64Main by getting {
+    dependencies {
+        dependsOn(iosMain.get())
+        kotlin.srcDir("build/generated/ksp/iosX64/iosX64Main/kotlin")
```
evant, line 55: Do we still need to add these as srcDirs? our multiplatform sample doesn't seem to have it https://github.com/evant/kotlin-inject-samples/blob/main/multiplatform/greeter/shared/build.gradle.kts, in theory it should be no longer necessary as of [KSP 1.8.0-1.0.9](https://kotlinlang.org/docs/ksp-quickstart.html#make-ide-aware-of-generated-code), but I haven't taken the time to verify.

evant (commented):
sorry, took a while to get to, overall I like the idea of adding this to the docs

docs/multiplatform.md
RivuChk: I needed this, compiler wasn't detecting the generated folder without this.
I tested it in 3 different projects (one where I'm actually using it on prod in JioCinema app, another one is in the courses I do on KMP, and lastly in the demo project for this document)

docs/multiplatform.md
RivuChk: Except adding the generated folders to sourcesets, the steps remain same if you add js to the target platform as well. I think I called that out as well.
However, I'm completely fine saying it's mostly towards Android/iOS

evant: Thinking something like "below is an example of an android/iOS multiplatform project, other targets follow the same pattern"

docs/multiplatform.md
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+```
+
+Add the compiler dependency in root level `dependencies` block for each of the platforms you're targetting
+```kotlin
+dependencies {
+    add("kspIosX64", "me.tatarka.inject:kotlin-inject-compiler-ksp:0.6.3")
+    add("kspIosArm64", "me.tatarka.inject:kotlin-inject-compiler-ksp:0.6.3")
+    add("kspIosSimulatorArm64", "me.tatarka.inject:kotlin-inject-compiler-ksp:0.6.3")
```
humblerookie, line 77: we might need one for `kspCommonMainMetadata` as well I suppose?

RivuChk: I didn't need that

docs/multiplatform.md
evant: I've updated the greeter sample and the IDE is resolving references fine without these additional srcDir declarations. Removing them from the doc but let me know if there's something I'm missing.

evant:
Took the liberty of making some changes and merged, feel free to suggest changes through another issue or pr if you feel like there's something I missed!

Issue #261, opened 2023-04-25
Allow to use "fun" as a "builder" of assisted dependency


dimsuz:
It seems that currently only having `val` lambdas is supported. This doesn't compile:

```kotlin
@Component
abstract class AssistedComponent {
  abstract fun target(arg: String): AssistedTarget
}

@Inject
class AssistedTarget(@Assisted arg: String)
```
failing with an error:
```
Class 'InjectAssistedComponent' is not abstract and does not implement 
abstract base class member public abstract fun target(arg: String)
```
Here's the generated class, which indeed lacks an "override":
```kotlin
public class InjectAssistedComponent() : AssistedComponent()
```

If I convert `fun` to `val`, it works without any problems and the correct code is generated. I.e. this works OK:
```kotlin
@Component
abstract class AssistedComponent {
  abstract val target: (arg: String) -> AssistedTarget
}
```

* I understand that README only mentions `val`, but somehow I expected that `fun` would've worked too, because it does work for non-assisted cases: I can use `val`/`fun` there as I please
* There are similar looking issues reported, but not quite like this one, close this if it's a duplicate.

evant:
Had considered implementing this but not sure it pulls it's weight as a feature, since it would only be for components, you still need the lambda form for anywhere else you'd be injecting it. But maybe it's clear enough what it does to be harmless to have as a feature. Either way, the current error is not good though.

As a side note: fun does work here, though maybe not as you expected:
```kotlin
@Component
abstract class AssistedComponent {
  abstract fun target(): (String) -> AssistedTarget
}
```

dimsuz:
> you still need the lambda form for anywhere else you'd be injecting it

Perhaps this could be done automagically: I have fun in component, but whenever lambda is injected, `kotlin-inject` would derive lambda type from fun-type and properly inject.

But in general I agree that having lambda everywhere might be a simpler and consistent thing to do, if communicated clearly.

Issue #252, opened 2023-02-11
Support factory interfaces for function injection


evant:
Function injection is a useful feature especially with [compose](https://developer.android.com/jetpack/compose). However, because it only works with function types you lose support for named & optional arguments which can make the signature harder to read. The proposed solution is to also support factory interfaces which support these features.

Currently you can do:
```kotlin
typealias MyFunction = (String) -> String

@Inject
fun MyFunction(dep: Dep, @Assisted arg: String): String = ...
```

In addition you should be able to do:
```kotlin
interface MyFunction {
  operator fun invoke(arg: String = "default"): String
}

@Inject
fun MyFunction(dep: Dep, @Assisted arg: String): String = ...
```

which can be called as:
```kotlin
MyFunction()
MyFunction(arg = "value")
```

Alternative:

Since we know which arguments are assisted we could in theory generate this factory function ourselves. However due to processing limitations they wouldn't be able to support default arguments. It would also mean more generated code exposed throughout the codebase which it limited to `MyComponent::create` currently. For these reasons I propose going with the more verbose version above. This may be re-visited if/when we get a way to read default arguments.

sergeshustoff:
Disclaimer: I seem to be failing to see the usecase for function injection. It's likely that I miss something
But in case I'm not missing something there here is my case for deprecating the whole thing)

I see a bunch of problems with function injection in general:
- typealias is required for function to be discoverable by DI, but it generally feels off and when function is renamed the typealias is not (though the matching name is required)
- typealias goes against #253 in a way that it is kind of a qualifier for the function while we can't do that for classes anymore. So api get's inconsistent here. I know, it's far fetched as typealias here is not exactly that, but still.
- in a way typealias does for a function what `@Provides` binding does for a class, but it's external to the component.
- seems that it breaks when 2 different types of magic mix in #388 (compose and kotlin-inject). While it is a bug in ksp I'd expect similar problems not to be uncommon simply because mixing magic like that is dangerous.
- in case of receiverFun from test it looks like receiver works as an assisted parameter, but it's not (and can't be) marked like that.
    
        typealias receiverFun = String.(arg: NamedFoo) -> String
        
        @Inject
        fun String.receiverFun(dep: Foo, @Assisted arg: NamedFoo): String = this
    
And then wrapping the function in a class with inject constructor doesn't look like much more code while makes it much more understandable and clear. Plus this way there is no need for the assisted injection (in this case):

    @Inject
[...]

PR #341, opened 2024-01-29
Implement SAM support for assisted injection

### LINKED ISSUE #243: Allow use of `fun interface` for assisted injection
Maybe this is too dagger-like, but since `fun interface`s are functionally equivalent to a functional type it should be technically possible to allow them for assisted injection.

The main benefit is that it allows a `Factory` type to be nested within your injectable class, which isn't possible with a typealias.

e.g. 
```kotlin
@Inject class MyClass(
  private val paramA: ClassA,
  private val paramB: ClassB,
) {
  fun interface Factory {
    fun create(paramB: ClassB): MyClass
  }
}

// Usage elsewhere
@Inject class MyConsumer(
  private val myClassFactory: MyClass.Factory,
)
```

dellisd:
Closes #243

It might be a bit messy, but it effectively maps it to a special function type for code generation.

### DIFF
--- a/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
+++ b/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
@@ -195,11 +195,15 @@ abstract class AstType : AstElement() {
 
     abstract fun isPlatform(): Boolean
 
+    abstract fun isSamInterface(): Boolean
+
     abstract fun isFunction(): Boolean
 
     abstract fun isTypeAlias(): Boolean
 
     abstract fun resolvedType(): AstType
+
+    abstract fun samReturnType(): AstType
 
     @Suppress("NOTHING_TO_INLINE")
     inline fun isNotUnit() = !isUnit()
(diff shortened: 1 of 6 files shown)


evant:
Thanks for the pr! There's some api design decisions that need to be made before fully committing to this (see https://github.com/evant/kotlin-inject/issues/261 and https://github.com/evant/kotlin-inject/issues/252) so it'll take some time to review, but I'll take this as a starting point.

ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
```diff
@@ -195,12 +195,16 @@ abstract class AstType : AstElement() {
...
+    abstract fun isSamInterface(): Boolean
+
     abstract fun isFunction(): Boolean
 
     abstract fun isTypeAlias(): Boolean
 
     abstract fun resolvedType(): AstType
 
+    abstract fun samReturnType(): AstType
+
```
evant, line 207: I think it would make the impl less awkward if we had

```
abstract fun samFunction(): AstFunction?
```

dellisd: I cleaned it up a little bit. I'm less sure about removing `isSamInterface` since it's a bit clearer when calling than checking if the result of `samFunction` is null.

evant:
Superseded by https://github.com/evant/kotlin-inject/pull/413

PR #343, opened 2024-01-30
Consider covariance when collecting type info

### LINKED ISSUE #342: Search for a covariant type when checking if a component's abstract property has been implemented
The following code causes an error:

> Cannot find an @Inject constructor or provider for: AppComponent

```kotlin
interface AppComponent

abstract class SessionComponent {
    abstract val appComponent: AppComponent
}

@Component
abstract class InheritedAppComponent : AppComponent

@Component
abstract class InheritedSessionComponent(
    @Component override val appComponent: InheritedAppComponent,
) : SessionComponent()
```

It looks like #326 was too aggressive in fixing #309

eygraber:
Fixes #342

### DIFF
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
@@ -280,7 +280,9 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
             val methods = mutableListOf<AstMember>()
             val isProvides = mutableMapOf<AstMember, Boolean>()
             for (method in allMethods) {
-                val existing = methods.firstOrNull { it.name == method.name && it.signatureEquals(method) }
+                val existing = methods.firstOrNull {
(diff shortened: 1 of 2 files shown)


kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
```diff
@@ -280,7 +280,9 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
             val methods = mutableListOf<AstMember>()
             val isProvides = mutableMapOf<AstMember, Boolean>()
             for (method in allMethods) {
-                val existing = methods.firstOrNull { it.name == method.name && it.signatureEquals(method) }
+                val existing = methods.firstOrNull {
+                    it.name == method.name && (it.signatureEquals(method) || it.overrides(method))
```
eygraber, line 284: I was thinking of doing something with `KsTypeReference.resolve.isAssignableFrom` but that didn't fit in `signatureEquals` (since it's technically not equal) and `overrides` was the previous behavior.

evant (approved):

PR #349, opened 2024-02-07
Prepare for K2

eygraber:
Was investigating using KSP 2 and figured I'd start tracking this separate from that.

### DIFF
--- a/kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
+++ b/kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
@@ -1,13 +1,14 @@
 package me.tatarka.inject.internal
 
-import kotlinx.atomicfu.locks.ReentrantLock
+import kotlinx.atomicfu.locks.reentrantLock
 import kotlinx.atomicfu.locks.withLock
 
 private val NULL = Any()
 
 actual class LazyMap {
     private val map = mutableMapOf<String, Any>()
-    private val lock = ReentrantLock()
+
+    private val lock = reentrantLock()
 
     actual fun <T> get(key: String, init: () -> T): T {
         return lock.withLock {
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,3 +2,4 @@ org.gradle.parallel=true
 kotlin.mpp.stability.nowarn=true
 kotlin.native.ignoreDisabledTargets=true
 org.gradle.jvmargs=-Xmx4G
+ksp.useKSP2=false
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,10 +1,10 @@
 [versions]
 kotlin-inject = "0.7.2-SNAPSHOT"
-kotlin = "1.9.24"
-ksp = "1.9.24-1.0.20"
+kotlin = "2.0.0"
+ksp = "2.0.0-1.0.22"
 kotlinpoet = "1.16.0"
 junit5 = "5.9.3"
-detekt = "1.23.4"
+detekt = "1.23.6"
 [libraries]
 kotlin-gradle = { module = "org.jetbrains.kotlin:kotlin-gradle-plugin", version.ref = "kotlin" }
 detekt-gradle = { module = "io.gitlab.arturbosch.detekt:detekt-gradle-plugin", version.ref = "detekt" }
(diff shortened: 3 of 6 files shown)


kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
```diff
@@ -1,13 +1,13 @@
...
-import kotlinx.atomicfu.locks.ReentrantLock
+import kotlinx.atomicfu.locks.SynchronizedObject
 import kotlinx.atomicfu.locks.withLock
 
 private val NULL = Any()
 
 actual class LazyMap {
     private val map = mutableMapOf<String, Any>()
-    private val lock = ReentrantLock()
+    private val lock = SynchronizedObject()
```
eygraber, line 10: https://github.com/Kotlin/kotlinx-atomicfu/issues/401

kotlin-inject-compiler/test/src/main/kotlin/me/tatarka/inject/ProjectCompiler.kt
```diff
@@ -43,6 +43,7 @@ class ProjectCompiler(
     fun compile(): TestCompilationResult {
         val result = TestCompilationResult(
             KotlinCompilation().apply {
+                kotlincArguments = kotlincArguments + "-Xskip-prerelease-check"
```
eygraber, line 46: Can be removed once K2 is stable

eygraber:
The integration tests got kind of screwy.

The issue is that currently KSP is configured to generate code for each platform. It is also configured to generate code for `commonMain`, however there is no code in commonMain because `common-companion` is a `commonTest` source set, and so there is no common code generated.

[In K2, common source sets can't see code from platform source sets](https://kotlinlang.slack.com/archives/C013BA8EQSE/p1707330548679529?thread_ts=1706907674.870889&cid=C013BA8EQSE), so the tests in `common` and `common-companion` don't see the generated platform code.

> It's expected that common code cannot reference generated code in the compilation of platform code. Generated codes are treated as platform code and K2 explicitly disallow references from common to platform (you'll have to use expect/actual).

To get around that I separated `common` and `common-companion` into the common code from them which becomes a `commonMain` source set, and their tests, which remain a `commonTest` source set.

Because it's using `commonTest` I believe tests still run for each platform, it's just that the generated code will be in the `commonMain` source set.

The only other solution that would work would be to have code and tests duplicated for each platform. You can't make multiple source sets out of the same directory, so there would probably have to be a script that generates them at runtime, or do some kind of trickery with symlinks.

kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
evant: is `SynchronizedObject` reentrant? Since we call into client code inside the lock I think it's very possible. Don't think we have a good test for this unfortunately.

evant:
The integration test setup is known to be kludge, it's only set up that way because I couldn't find another way to make it work.

> Because it's using commonTest I believe tests still run for each platform, it's just that the generated code will be in the commonMain source set.

👍  That feels right to me and closer to the ideal way I wanted it set up.

kotlin-inject-runtime/src/nativeMain/kotlin/me/tatarka/inject/internal/LazyMap.kt
evant: Looks like yes:
https://github.com/Kotlin/kotlinx-atomicfu/commit/3b895fe0409d5195877d813581edeb88cf014774#diff-3a82b7dd4d92bba2b14c26b50c02172b206613984ac478fc4d29b432f508f88cR29

https://docs.oracle.com/javase/tutorial/essential/concurrency/locksync.html

eygraber: I believe `ReentrantLock` is just a typealias for `SynchronizedObject` on Native.

eygraber:
@evant are you OK with extracting the integration test changes and merging into `main` now? It should work fine on current `main` and would prevent conflicts as new tests are added.

evant:
I was just going to ask to do that, these changes are quite large and it would be easier to review if the integration test changes were pulled out

eygraber:
https://github.com/evant/kotlin-inject/pull/355

evant:
I think I want to get a release out with the current changes before merging this but happy to do right after

evant:
Released, so this can be moved out of draft

evant (approved):

PR #350, opened 2024-02-08
Use KSP2

eygraber:


kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/Util.kt
```diff
@@ -35,7 +35,7 @@ private fun KSAnnotation.hasName(packageName: String, simpleName: String): Boole
     // we can skip resolving if the short name doesn't match
     if (shortName.asString() != simpleName) return false
     val declaration = annotationType.resolve().declaration
-    return declaration.packageName.asString() == packageName
+    return declaration.qualifiedName?.asString() == "$packageName.$simpleName"
```
eygraber, line 38: https://github.com/google/ksp/issues/1721

evant: the upstream bug is fixed now?

kotlin-inject-compiler/test/src/main/kotlin/me/tatarka/inject/ProjectCompiler.kt
```diff
@@ -107,5 +107,6 @@ class TestCompilationResult(private val result: CompilationResult) {
     val success: Boolean
         get() = result.exitCode == KotlinCompilation.ExitCode.OK
 
+    @OptIn(ExperimentalCompilerApi::class)
```
evant, line 110: compile-testing should have a way to filter output in 0.5.0 that can replace our custom code https://github.com/ZacSweers/kotlin-compile-testing/issues/198#issuecomment-1931317732

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/Util.kt
eygraber: I think I tried it earlier and it still had issues, but that might have been conflated with the other KSP2 issues. I'll try again.

eygraber: I reverted the commit so it could be unreverted in case it is still needed.

kotlin-inject-compiler/test/src/main/kotlin/me/tatarka/inject/ProjectCompiler.kt
eygraber: I'm not sure if that would be a better solution here. [I pushed a commit showing the changes](https://github.com/evant/kotlin-inject/pull/350/commits/bb532681c88e88566193b513292da8734899443a), but the issue is that you can't filter after the fact. You have to specify in the `KotlinCompilation` what the logging level should be, which means you need to pass that to `ProjectCompiler.compile`.

evant: True, just left a comment about this. Can wait to see if there's any desire to change anything upstream, but otherwise would prefer to get rid of our code even if it's not quite as nice.

evant: Ok, made a [pr](https://github.com/ZacSweers/kotlin-compile-testing/pull/260) upstream, can leave it how it is for now.

eygraber: Pushed a commit addressing this

eygraber:
@evant I updated all dependencies to latest versions, but looks like https://github.com/google/ksp/issues/1854 is still an issue, as well as some issues with `typealiases` which might be related to https://kotlinlang.slack.com/archives/C013BA8EQSE/p1716431911867909. I tested with 2.1.0-RC and both issues are present there as well.

I probably won't have much time to dedicate to this for the next few months, in case there's anyone who'd like to pick it up.

ting-yuan:
The exception `KtInvalidLifetimeOwnerAccessException: Access to invalid KtAlwaysAccessibleLifetimeToken: PSI has changed since creation`, was thrown in the second round in the `SymbolProcessor.finish()`
```
invoke:79, AbstractKSDeclarationImpl$containingFile$2 (com.google.devtools.ksp.impl.symbol.kotlin)
invoke:78, AbstractKSDeclarationImpl$containingFile$2 (com.google.devtools.ksp.impl.symbol.kotlin)
lambda 'synchronized' in 'value':74, SynchronizedLazyImpl (kotlin)
getValue:69, SynchronizedLazyImpl (kotlin)
getContainingFile:78, AbstractKSDeclarationImpl (com.google.devtools.ksp.impl.symbol.kotlin)
lambda 'apply' in 'addOriginatingElement':95, KSAstProvider (me.tatarka.kotlin.ast)
addOriginatingElement:94, KSAstProvider (me.tatarka.kotlin.ast)
lambda 'forEach' in 'generate':27, KmpComponentCreateGenerator (me.tatarka.inject.compiler)
forEach:1863, KmpComponentCreateGenerator (me.tatarka.inject.compiler)
lambda 'apply' in 'generate':23, KmpComponentCreateGenerator (me.tatarka.inject.compiler)
lambda 'with' in 'generate':22, KmpComponentCreateGenerator (me.tatarka.inject.compiler)
generate:18, KmpComponentCreateGenerator (me.tatarka.inject.compiler)
lambda 'forEach' in 'generateKmpComponentCreateFiles':40, ProcessKmpComponentCreateKt (me.tatarka.inject.compiler.ksp)
forEach:216, ProcessKmpComponentCreateKt (me.tatarka.inject.compiler.ksp)
generateKmpComponentCreateFiles:39, ProcessKmpComponentCreateKt (me.tatarka.inject.compiler.ksp)
finish:113, InjectProcessor (me.tatarka.inject.compiler.ksp)
lambda 'forEach' in 'execute':586, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
execute$stub_for_inlining$32:1863, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
forEach:1863, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
execute:586, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
...
```

whereas the corresponding element of `createNative` was from the first round:
```
lambda 'with' in 'processKmpComponentCreate':29, ProcessKmpComponentCreateKt (me.tatarka.inject.compiler.ksp)
processKmpComponentCreate:16, ProcessKmpComponentCreateKt (me.tatarka.inject.compiler.ksp)
process$lambda$3:69, InjectProcessor (me.tatarka.inject.compiler.ksp)
invoke:-1, InjectProcessor$$Lambda/0x00007f43f947f670 (me.tatarka.inject.compiler.ksp)
calcNext:171, FilteringSequence$iterator$1 (kotlin.sequences)
hasNext:194, FilteringSequence$iterator$1 (kotlin.sequences)
toList:813, SequencesKt___SequencesKt (kotlin.sequences)
process:70, InjectProcessor (me.tatarka.inject.compiler.ksp)
invoke:552, KotlinSymbolProcessing$execute$1$1 (com.google.devtools.ksp.impl)
invoke:550, KotlinSymbolProcessing$execute$1$1 (com.google.devtools.ksp.impl)
closeFilesOnException:409, IncrementalContextBase (ksp.com.google.devtools.ksp.common)
lambda 'forEach' in 'execute':550, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
forEach:1863, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
execute:549, KotlinSymbolProcessing (com.google.devtools.ksp.impl)
...
```

Please avoid caching and using the elements across rounds.

evant:
@eygraber ^ probably due to https://github.com/evant/kotlin-inject/blob/7f7dcd920b5ca792dce5f62199d36d389ac3c548/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt#L35 we should change it to only store the `KSName` like we do for other deferred symbols.

eygraber:
Yup, I started working on that. Not sure if I'm doing something wrong because I'm getting other errors but at least the `KtInvalidLifetimeOwnerAccessException` seems to have stopped :smile:

eygraber:
@evant I believe the latest commit fixes the `KtInvalidLifetimeOwnerAccessException` issue.

There is still an issue with typealiases (probably related to https://github.com/google/ksp/issues/1921).

Most of the `RoundsTests` are failing because of a change in `dev.zacsweers.kctfork:ksp:0.6.0` where the messages are split between a `messages: String` property, and a `diagnosticMessages: List<DiagnosticMessage>`. `diagnosticMessages` is used in `messagesWithSeverity` (which is what we call to get the error output), however the errors that we are looking for only appear in `messages`.

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/FailureTest.kt
```diff
@@ -1495,7 +1495,7 @@ class FailureTest {
                 """.trimIndent()
             ).compile()
         }.output().all {
-            contains("e: [ksp] Cannot find an @Inject constructor or provider for: Foo")
+            contains("Cannot find an @Inject constructor or provider for: Foo")
```
eygraber, line 1498: @evant The `e: [ksp]` prefix wasn't present anymore; it was a `toString` representation of `DiagnosticMessage`

eygraber:
I'm hoping this is "good to go" (probably could use some cleanup though). Some notes:

1. I basically reverted https://github.com/evant/kotlin-inject/pull/350/commits/ba386e53a6145986aa8762120c44ba509b13d720 because of the issues with `diagnosticMessages` that I commented about above. I tried debugging that for a while, and I think it has something to do with the `MessageCollector` from `AbstractKotlinCompilation.createMessageCollector` not being the `MessageCollector` that gets used internally, but I wanted to move forward
2. A lot of the tests are failing when reverting to use KSP1. I don't know what the impact of that is for end users.
3. I'm concerned there are more typealias issues lurking that haven't been exposed by tests yet
4. The `Ast` abstraction is nice in theory, but it makes debugging much harder

eygraber:
Also I've seen the current CI failure locally, and I get rid of it by cleaning and running again. Maybe circleci caches need to be cleaned?

vRallev:
The workaround for #447 isn't user friendly.

eygraber:
Agreed, I'm updating the issue you filed on KSP.

evant:
> Maybe circleci caches need to be cleaned?

you can invalidate the cache by changing the cache_key at https://github.com/evant/kotlin-inject/blob/7f7dcd920b5ca792dce5f62199d36d389ac3c548/.circleci/config.yml#L62 & https://github.com/evant/kotlin-inject/blob/7f7dcd920b5ca792dce5f62199d36d389ac3c548/.circleci/config.yml#L82

btw thanks @eygraber for keeping up with this

eygraber:
It looks like this happens when running tests after assemble. I'll dig in further.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/KmpComponentCreateGenerator.kt
```diff
@@ -1,20 +1,39 @@
...
+
+data class KmpComponentCreateFunctionInfo(
+    val name: String,
+    val annotations: List<AnnotationSpec>,
+    val visibility: KModifier,
+    val receiverParameterType: TypeName?,
+    val parameters: List<Pair<String, TypeName>>,
+    val parametersTemplate: String,
+    val parameterNames: List<String>,
+)
```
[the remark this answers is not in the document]
eygraber: I agree. The fastest way I could think of to do this was to defer all functions found with `resolver.getSymbolsWithAnnotation` and then call `resolver.getSymbolsWithAnnotation` in `finish` and generate everything there.

Not sure if that's frowned upon in KSP, so I'm also working on a solution that generates a file per function so that nothing needs to be cached. The difficulty there is naming the generated file because it's easy to end up with multiple files named the same.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/KmpComponentCreateGenerator.kt
```diff
@@ -24,36 +43,37 @@ class KmpComponentCreateGenerator(
                 addFunction(
                     FunSpec
                         .builder(kmpComponentCreateFunction.name)
-                        .addOriginatingElement(kmpComponentCreateFunction)
+//                        .addOriginatingElement(kmpComponentCreateFunction)
```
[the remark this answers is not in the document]
eygraber: Mistake :sweat_smile:

eygraber:
CI is blocked by https://github.com/google/ksp/issues/2230. Alternatively we can update the script to run `clean` in between `assemble` and `check`, and revert that once the KSP issue is resolved.

Outstanding questions are:

1. Is it OK if kotlin-inject doesn't work with KSP1 anymore?
2. Is it OK to defer all KmpComponentCreate functions to the last round, or is a different solution needed?

evant:
> Is it OK if kotlin-inject doesn't work with KSP1 anymore?

I've seen enough issues that I think it would be good to have overlap supporting both for a bit. What breaks here with ksp1? I don't know if it'll easy to detect which one you are using but we could add an additional option to enable it for now

eygraber:
Actually it looks like it does work with KSP1. The integration tests are all passing, but about half of the unit tests fail. Maybe I'm not configuring `kotlin-compile-testing` correctly, or maybe the tests are actually failing, it's hard to say for sure.

eygraber:
Got some inspiration from https://github.com/amzn/kotlin-inject-anvil/blob/main/compiler-utils/src/testFixtures/kotlin/software/amazon/lastmile/kotlin/inject/anvil/Compilation.kt and figured it out :smile: 

The only unit tests failing now are the ones that check for a specific error format that changed from KSP1 to KSP2, so I'm thinking that it is fine.

eygraber:
Updated the unit and integration tests to run with both KSP1 and KSP2.

evant:
> CI is blocked by https://github.com/google/ksp/issues/2230

looks like it's already fixed! I'm ok with throwing a clean on ci for now until that's released though

eygraber:
Updated, hopefully that's all that's needed.

sahilbajaj:
@eygraber @evant Eagerly awaiting action.

eygraber:
@evant anything still needed here (aside from getting past New Years)?

evant:
Will take a final pass when I get a chance but don't imagine anything major blocking merge, no

eygraber:
It looks like function injection isn't working where the typealias points to another typealias. I don't have the time to investigate now, but that wasn't working before these changes either, so I'm guessing not a blocker?

evant:
> so I'm guessing not a blocker

yeah, if it's not a regression it's not a blocker

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/Util.kt
```diff
@@ -72,4 +72,15 @@ private fun <T> renderTree(node: T, visitor: TreeVisitor<T>, indent: String, isL
 
 fun AstType.isSet(): Boolean = packageName == "kotlin.collections" && simpleName == "Set"
 fun AstType.isMap(): Boolean = packageName == "kotlin.collections" && simpleName == "Map"
-fun AstType.isPair(): Boolean = packageName == "kotlin" && simpleName == "Pair"
\ No newline at end of file
+fun AstType.isPair(): Boolean = packageName == "kotlin" && simpleName == "Pair"
+fun AstType.isFunctionOrTypeAliasOfFunction(): Boolean = isFunction() || isTypeAlias() &&
+    resolvedType().isFunctionOrTypeAliasOfFunction()
+
+tailrec fun AstType.resolveToHighestTypeAlias(): AstType {
```
evant, line 79: This always returns the concrete type correct? I think the name is a bit confusing, maybe something like `fullyResolvedType()`? This could also go on the ast class instead of as an extension but don't have strong opinions about that.

kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/InheritanceTest.kt
```diff
@@ -2,20 +2,22 @@ package me.tatarka.inject.test
...
 import org.junit.jupiter.api.assertDoesNotThrow
 import org.junit.jupiter.api.io.TempDir
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.EnumSource
 import java.io.File
 
 class InheritanceTest {
 
-    private val target = Target.KSP
+    private val target = Target.KSP2
```
evant, line 13: unused

evant (approved):

eygraber:
@evant with KSP 2.0.0 enabling KSP2 by default, can we get a release that includes this?

PR #357, opened 2024-02-15
Consider Scope annotations from superclasses

### LINKED ISSUE #354: Scope annotations aren't picked up when scanning superclasses
In the following case `FooSingleton` isn't picked up from the `BaseFooComponent` interface:

```kotlin
@Scope
annotation class FooSingleton

interface Foo

@Inject
class MyFoo : Foo

@Inject
class FooHolder(
  private val foo: Foo
)

interface BaseFooComponent {
  @Provides @FooSingleton fun provideFoo(): Foo
}

@FooSingleton
@Component
abstract class FooComponent : BaseFooComponent {
  override fun provideFoo(): Foo = MyFoo()

  abstract val holder: FooHolder

  companion object
}
```

The following is generated, without `provideFoo` being scoped:

```kotlin
public fun FooComponent.Companion.create(): FooComponent = InjectFooComponent()

public class InjectFooComponent() : FooComponent(), ScopedComponent {
  override val _scoped: LazyMap = LazyMap()

  override val holder: FooHolder
    get() = FooHolder(
      foo = provideFoo()
    )
}
```

Adding `@FooSingleton` to `provideFoo` in `FooComponent` does cause it to get scoped:

```kotlin
public fun FooComponent.Companion.create(): FooComponent = InjectFooComponent()

public class InjectFooComponent() : FooComponent(), ScopedComponent {
  override val _scoped: LazyMap = LazyMap()

  override val holder: FooHolder
    get() = FooHolder(
      foo = _scoped.get("com.example.Foo") {
        provideFoo()
      }
    )
}
```

eygraber:
Fixes #354

### DIFF
--- a/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
+++ b/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
@@ -107,8 +107,8 @@ private interface KSAstAnnotated : AstAnnotated {
         return declaration.hasAnnotation(packageName, simpleName)
     }
 
-    override fun annotationAnnotatedWith(packageName: String, simpleName: String): AstAnnotation? {
-        return declaration.annotationAnnotatedWith(packageName, simpleName)?.let {
+    override fun annotationsAnnotatedWith(packageName: String, simpleName: String): Sequence<AstAnnotation> {
+        return declaration.annotationsAnnotatedWith(packageName, simpleName).map {
             KSAstAnnotation(resolver, it)
         }
     }
@@ -362,9 +362,15 @@ private class KSAstProperty(override val resolver: Resolver, override val declar
         return declaration.getter?.hasAnnotation(packageName, simpleName) == true
     }
 
-    override fun annotationAnnotatedWith(packageName: String, simpleName: String): AstAnnotation? {
-        return declaration.getter?.annotationAnnotatedWith(packageName, simpleName)?.let {
-            KSAstAnnotation(resolver, it)
+    override fun annotationsAnnotatedWith(packageName: String, simpleName: String): Sequence<AstAnnotation> {
+        val declarationAnnotations = super.annotationsAnnotatedWith(packageName, simpleName)
+        val getter = declaration.getter
+        return if (getter == null) {
+            declarationAnnotations
+        } else {
+            getter.annotationsAnnotatedWith(packageName, simpleName).map { annotation ->
+                KSAstAnnotation(resolver, annotation)
+            } + declarationAnnotations
         }
     }
 
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
@@ -99,8 +99,7 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
                 }
             }
 
-            for (method in typeInfo.providesMethods) {
-                val scope = method.scopeType(options)
+            for ((method, scope) in typeInfo.providesMethods) {
                 if (scope != null && scope != typeInfo.elementScope) {
                     if (typeInfo.elementScope != null) {
                         provider.error(
@@ -253,7 +252,7 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
         return typeInfoCache.getOrPut(astClass.toString()) {
             val isComponent = astClass.isComponent()
 
-            val providesMethods = mutableListOf<AstMember>()
+            val providesMethods = mutableListOf<Pair<AstMember, AstType?>>()
             val providerMethods = mutableListOf<AstMember>()
 
             var scopeClass: AstClass? = null
@@ -279,7 +278,10 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
             // some methods may override others
             val methods = mutableListOf<AstMember>()
             val isProvides = mutableMapOf<AstMember, Boolean>()
+            val scopeTypes = mutableMapOf<AstMember, Set<AstType>>()
+            val scopedMethodsWithScopedSuperMethod = mutableMapOf<AstMember, Pair<AstMember, Set<AstType>>>()
             for (method in allMethods) {
+                val methodScopeTypes = method.scopeTypes(options).toSet()
                 val existing = methods.firstOrNull {
                     it.name == method.name && (it.signatureEquals(method) || it.overrides(method))
                 }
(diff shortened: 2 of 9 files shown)


ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
```diff
@@ -110,7 +110,23 @@ private interface KSAstAnnotated : AstAnnotated {
     override fun annotationAnnotatedWith(packageName: String, simpleName: String): AstAnnotation? {
         return declaration.annotationAnnotatedWith(packageName, simpleName)?.let {
             KSAstAnnotation(resolver, it)
-        }
+        } ?: overrideeAnnotationAnnotatedWith(declaration, packageName, simpleName)
```
evant, line 113: hm, I wonder if doing this here would be confusing, especially since we are explicitly looking for overrides in other places in the code. I was thinking of doing it in [TypeCollector](https://github.com/evant/kotlin-inject/blob/81280fb80e6fd2040127e649d4afa267ad76e1bc/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt) which is where we are collecting provides methods. It could check for scopes as well when checking for `@Provides` on the overriden method in [collectTypeInfo()](https://github.com/evant/kotlin-inject/blob/81280fb80e6fd2040127e649d4afa267ad76e1bc/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt#L282) and use that information for obtaining the scope in [collectTypes()](https://github.com/evant/kotlin-inject/blob/81280fb80e6fd2040127e649d4afa267ad76e1bc/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt#L103)

eygraber: I was going to ask about that (but got distracted making it work for properties). Unless there's certain cases where we don't want to look at super types, maybe it makes sense to just do it here?

evant: Hm, wouldn't be against it. But would want to change how `collectTypeInfo` works then so we aren't double-checking. And write a few more tests to make sure we didn't break anything.

evant: Er, also want to note that we need to scan superclass methods anyway regardless on if they override or not. The use of a hashmap for checking overrides was to prevent causing an n^2 lookup, though haven't check if `findOverridee()` does something smarter than linearly scanning methods for a match.

eygraber: I'm looking into it now, but the way that `@Provides` is currently handled might be the better approach. Also the docs for `findOverridee()` mention that:

> Calling `findOverridee` is expensive and should be avoided if possible

eygraber: Does it make sense for a `@Provides` or a class to have multiple scope annotations? If not, I think a more up front error could be clearer here.

The current behavior is that the first scope annotation is used, which leads to failures later if the first scope doesn't match the scope of the component that it is being used with.

evant: we error if there's conflicting scopes on the class, I think we should error in the same way here

evant: I think an error would be easier to understand there, we could always relax later if there's a real need for that pattern.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
```diff
@@ -253,7 +252,7 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
         return typeInfoCache.getOrPut(astClass.toString()) {
             val isComponent = astClass.isComponent()
 
-            val providesMethods = mutableListOf<AstMember>()
+            val providesMethods = mutableListOf<Pair<AstMember, AstType?>>()
```
eygraber, line 255: @evant would you prefer a custom type instead of `Pair`?

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
```diff
@@ -279,7 +278,10 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
             // some methods may override others
             val methods = mutableListOf<AstMember>()
             val isProvides = mutableMapOf<AstMember, Boolean>()
+            val scopeTypes = mutableMapOf<AstMember, Set<AstType>>()
+            val scopedMethodsWithScopedSuperMethod = mutableMapOf<AstMember, Pair<AstMember, Set<AstType>>>()
```
eygraber, line 282: @evant would you prefer a custom type instead of Pair?

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
```diff
@@ -320,13 +356,12 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
                         provider.error("@Provides method must have a concrete implementation", method)
                         continue
                     } else {
-                        providesMethods.add(method)
+                        providesMethods.add(method to methodScope)
                     }
                 } else if (method.isProvider()) {
-                    val scope = method.scopeType(options)
-                    if (scope != null) {
+                    if (methodScopeTypes != null) {
```
evant, line 362: should this be `isNotEmpty()`?

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
evant: no strong opinions

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
evant: no strong opinions

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
eygraber: This case is different because it is pulling it out of the `scopeTypes` Map on line 314

eygraber: It's a little confusing because the invariant is that if there is a `Set<AstType>` in `scopeTypes` then it won't be empty.

evant: ah looking at the full context I understand

evant (approved):

PR #362, opened 2024-02-21
Add an Option to generate actual create functions

### LINKED ISSUE #361: Expect Actual Ancestor Components and Common Descendant Components
I have a component hierarchy that has an expect component in the middle of it, and the actual implementations provide platform specific bindings. That requires that all downstream components be generated for all platforms, even if they themselves are common, because common generated code wouldn't be able to see these platform specific bindings.

Pre Kotlin 2.0 this isn't a problem in some cases because common source sets can (incorrectly) reference generated platform code. However once Kotlin 2.0 is released this won't work anymore for any of the cases.

To solve the issue, all downstream components (which is ~99% of the components in my project) will need to specify expect/actual declarations for the `create` functions for each platform (this project supports 5 platform targets) in each module (and I typically have 1 downstream component per module).

That's kind of painful, and I'm wondering if there's a way for kotlin-inject to detect (or be told about) this situation, and generate the expect/actual declarations for `create`.

I'm not sure if it could be done in the same processor, or if it would need a different processor, or if it's even possible with KSP at all (I think it would need something like [Support source set-specific code generation in Multiplatform projects](https://github.com/google/ksp/issues/965)).

Other options that (hopefully) don't require changes in KSP are:

1. A flag for kotlin-inject that will mark the generated `create` function as actual and then the user just needs to specify a common `expect create` function
2. A new annotation can be introduced to use with descendant components that live in common, and the presence of that annotation will:
   - generate a common `expect create` function
   - mark generated platform `create` functions as actual

eygraber:
Fixes #361


integration-tests/ksp-kmp/app/build.gradle.kts
eygraber, line 1: I can extract this into `build-logic` as `kotlin-inject.multiplatform-tests.gradle.kts`. The idea is to not have to provide actual implementations for each target, so it just uses a subset (i.e. jvm and js).

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/Options.kt
```diff
@@ -4,19 +4,22 @@ private const val OPTION_GENERATE_COMPANION_EXTENSIONS = "me.tatarka.inject.gene
 private const val OPTION_ENABLE_JAVAX_ANNOTATIONS = "me.tatarka.inject.enableJavaxAnnotations"
 private const val OPTION_USE_CLASS_REFERENCE_FOR_SCOPED_ACCESS = "me.tatarka.inject.useClassReferenceForScopeAccess"
 private const val OPTION_DUMP_GRAPH = "me.tatarka.inject.dumpGraph"
+private const val OPTION_GENERATE_ACTUAL_CREATE_FUNCTION = "me.tatarka.inject.generateActualCreateFunction"
```
eygraber, line 7: The naming is not great because "actual" is ambiguous, and can be read a few different ways.

eygraber:
@evant any thoughts on this?

eygraber:
I added a processor to my project that generates actual functions that call the target create function - https://github.com/eygraber/immich-kmp/blob/e5aa43231926962dd23f228a55bcd49514e7cd2e/ksp/generate-actual-compiler/src/main/kotlin/app/immich/kmp/ksp/generate/actual/GenerateProcessorProvider.kt

The usage looks like:

```kotlin
@GenerateActual
internal expect fun MainHostComponent.Companion.createKmp(
  sessionComponent: ImmichSessionComponent,
  route: Route,
): MainHostComponent
```
which generates:

```kotlin
@GenerateActual
internal actual fun MainHostComponent.Companion.createKmp(
    sessionComponent: ImmichSessionComponent,
    route: Route
): MainHostComponent = MainHostComponent.Companion.create(sessionComponent, route)
```

evant:
sorry taking a long time to reply, this really feels like a giant hack which makes me hesitant to do something about it. I do like your additional processor solution better than an extra flag though.

eygraber:
I can make a new PR using the additional processor idea.

eygraber:
Any thoughts on what the new annotation could be called? I used `GenerateActual` as a placeholder, but I don't think that's good to release.

eygraber:
Closing because of https://github.com/evant/kotlin-inject/pull/370

PR #370, opened 2024-03-27
Add a new annotation to generate Component accessors for KMP targets

### LINKED ISSUE #361: Expect Actual Ancestor Components and Common Descendant Components

### LINKED ISSUE #371: Component with an Int literal property name is used as an Int value
The following code uses the property name as a value:

```kotlin
@Suppress("LocalVariableName")
@Component
abstract class MyComponent(
    `1`: String
)
```

generates:

```kotlin
public fun KClass<MyComponent>.create(
  `1`: String
): MyComponent = InjectMyComponent(1)

public class InjectMyComponent(
  `1`: String
) : MyComponent(`1`)
```

eygraber:
Fixes #361 
Fixes #371


eygraber:
@evant naming is still not great, but I think it's kind of descriptive of what it is doing.

What do you think in terms of the module location, and moving the KSP test harness so that it could be used by both KSP modules?

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
```diff
@@ -0,0 +1,146 @@
...
+            if (symbol !is KSFunctionDeclaration) {
+                logger.error("TargetComponentAccessor should be applied to a function")
+                continue
+            }
+
+            val funcName = symbol.simpleName.asString()
+
+            // the generated actual funs will be annotated with TargetComponentAccessor
+            // KSP will process them as well so we need to ignore them
+            if (Modifier.ACTUAL in symbol.modifiers && symbol.findExpects().firstOrNull() != null) {
```
evant, line 44: > the generated actual funs will be annotated with TargetComponentAccessor

does it need to be?

settings.gradle.kts
```diff
@@ -29,3 +29,4 @@ include(":integration-tests:ksp")
 include(":integration-tests:module")
 include(":integration-tests:ksp-companion")
 include(":integration-tests:jmh")
+include(":target-component-accessor-compiler:ksp")
```
evant, line 32: seems like it would be easy to add the annotation but forget to include the extra compiler dependency. It's simple enough that maybe we don't need to split it out?

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
```diff
@@ -30,6 +30,7 @@ val INJECT = ClassName(ANNOTATION_PACKAGE_NAME, "Inject")
 val INTO_MAP = ClassName(ANNOTATION_PACKAGE_NAME, "IntoMap")
 val INTO_SET = ClassName(ANNOTATION_PACKAGE_NAME, "IntoSet")
 val ASSISTED = ClassName(ANNOTATION_PACKAGE_NAME, "Assisted")
+val TARGET_COMPONENT_ACCESSOR = ClassName(ANNOTATION_PACKAGE_NAME, "TargetComponentAccessor")
```
evant, line 33: is this used?

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
```diff
@@ -0,0 +1,146 @@
...
+
+class TargetComponentAccessorProcessor(
+    private val codeGenerator: CodeGenerator,
+    private val logger: KSPLogger,
+) : SymbolProcessor {
+    @Suppress("LoopWithTooManyJumpStatements")
+    override fun process(resolver: Resolver): List<KSAnnotated> {
+        val symbols = resolver.getSymbolsWithAnnotation(
+            requireNotNull(TargetComponentAccessor::class.qualifiedName),
+        )
```
evant, line 32: `getSymbolsWithAnnotation` does a pretty aggressive scan of all classes, methods, and nested classes. Would there be any downside to only scanning top-level functions? You can look at the existing processor to see how this can be done.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        return emptyList()
+    }
+}
+
+private fun KSFunctionDeclaration.validateReceiver(): KSTypeReference? {
+    val funcReceiver = extensionReceiver
+    val funcReceiverName = funcReceiver?.resolve()?.declaration?.qualifiedName?.asString()
+    return funcReceiver?.takeIf {
+        funcReceiverName != null && funcReceiverName.endsWith("Companion")
```
evant, line 82: seems like `endsWith("Companion")` is too loose? For example it'd match `MyCustomCompanion.foo()`.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+private fun KSFunctionDeclaration.validateReturnType(): KSDeclaration? {
+    return returnType?.resolve()?.declaration?.takeIf { funcReturn ->
```
evant, line 87: If the component itself is generated (say with some kind of anvil-like extension) then this may not resolve in the current round. You should check this and delay processing (see: https://kotlinlang.org/docs/ksp-multi-round.html)

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
```diff
@@ -0,0 +1,146 @@
...
+                    ),
+                )
+
+                // if we're at this point then returnType can't be null
+                returns(symbol.returnType!!.toTypeName())
+            }.build(),
+        )
+    }.build()
+
+    file.writeTo(codeGenerator, aggregating = true)
```
evant, line 138: why is this aggregating, there's a 1:1 relationship between source and generated no?

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
```diff
@@ -0,0 +1,146 @@
...
+    symbol: KSFunctionDeclaration,
+    name: String,
+    receiverType: KSTypeReference,
+    returnType: KSDeclaration,
+    codeGenerator: CodeGenerator
+) {
+    val file = FileSpec.builder(
+        packageName = symbol.packageName.asString(),
+        fileName = "ActualInject${returnType.simpleName.asString()}",
+    ).apply {
```
evant, line 103: you need to add the source file to the spec builder so it can be tracked for [incremental processing](https://square.github.io/kotlinpoet/interop-ksp/#incremental-processing).

kotlin-inject-runtime/src/commonMain/kotlin/me/tatarka/inject/annotations/Annotations.kt
```diff
@@ -26,4 +26,7 @@ annotation class IntoSet
 annotation class IntoMap
 
 @Target(VALUE_PARAMETER)
-annotation class Assisted
\ No newline at end of file
+annotation class Assisted
+
+@Target(FUNCTION)
+annotation class TargetComponentAccessor
```
evant, line 32: Finding a good name for this is tricky...

If my understanding is correct, the method is a way to reference the actual create function for a component from the shared source set.  How about `ReferenceActualComponent`?

evant (changes requested):
This will also needed tests. Not splitting out the processor into it's own module should make that easier to set them up.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
eygraber: I think it's currently a warning if it isn't but it will be turning into an error in k2

settings.gradle.kts
eygraber: I didn't do that because there could be problems if the processor is added to the common configuration, but maybe that's fine because using this annotation when the processor is in the common configuration is already a mistake. Hopefully the error (`actual` functions generated in the common source set) will be clear enough that the user realizes what mistake they made.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
eygraber: I was experimenting with a few things. I'll take it out if it doesn't end up getting used.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
eygraber: Yeah that should be fine.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
eygraber: Yeah, I just realized I could check the simple name because that would just be `Companion` in the correct case.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
eygraber: Just copy paste :smile:

kotlin-inject-runtime/src/commonMain/kotlin/me/tatarka/inject/annotations/Annotations.kt
eygraber: I was approaching it more as a way to access the create function that is generated in each target's source set from the shared source set. So it doesn't really reference it as much as call through to it.

target-component-accessor-compiler/ksp/src/main/kotlin/me/tatarka/inject/target/component/accessor/compiler/ksp/TargetComponentAccessorProcessor.kt
evant: oh if expect & actual don't match annotations? Yeah I guess that makes sense

eygraber: Yup

settings.gradle.kts
evant: gotcha, in that case failing loudly may actually be better than doing nothing assuming the error makes it obvious enough which sounds like it might

evant:
Oh one more thought, do we actually care about the companion here? Seems to me this is just a convenience in how the function is called, but

```kotlin
@TargetComponentAccessor
expect fun createMyCustomComponent(): MyCustomComponent
```

would work just as well

eygraber:
I was using the receiver to determine which create function to call, but I guess it makes more sense to use the return type. I also need to check if I should use the companion or KClass

evant:
You can check `me.tatarka.inject.generateCompanionExtensions` to determine if you are generating kclass or companion calls

eygraber:
Pushed all of those changes. I'll start writing tests tomorrow, but I did some refactoring to separate the Inject and TargetComponentAccessor processing, and wanted to get you thoughts on that.

evant (commented):
Have several nits but overall I like the approach 👍

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TargetComponentAccessorGenerator.kt
```diff
@@ -0,0 +1,70 @@
...
+                        for (param in astFunction.parameters) {
+                            addParameter(param.name, param.type.toTypeName())
+                        }
+
+                        val funcParams = astFunction.parameters.joinToString { it.name }
+
+                        val returnTypeCompanion = when {
+                            options.generateCompanionExtensions -> returnTypeClass.companion?.type
+                            else -> null
+                        }
```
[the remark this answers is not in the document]
eygraber: Yes, I already did that, but didn't push yet because there was a bug here

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TargetComponentAccessorGenerator.kt
```diff
@@ -0,0 +1,70 @@
...
+            packageName = astFunction.packageName,
+            fileName = "Target${returnTypeClass.name}Accessor",
+        ).apply {
+            addFunction(
+                FunSpec
+                    .builder(astFunction.name)
+                    .addOriginatingElement(astFunction)
+                    .apply {
+                        addAnnotation(TargetComponentAccessor::class)
+                        astFunction.optInAnnotation()?.let(::addAnnotation)
```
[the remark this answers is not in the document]
eygraber: I was thinking that, but didn't see a way to do it with Ast. I can add that.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/ProcessTargetComponentAccessor.kt
```diff
@@ -0,0 +1,60 @@
...
+internal fun processTargetComponentAccessor(
+    element: KSFunctionDeclaration,
+    provider: KSAstProvider,
+    codeGenerator: CodeGenerator,
+    targetComponentAccessorGenerator: TargetComponentAccessorGenerator,
+    skipValidation: Boolean = false,
+): Boolean = with(provider) {
+    val astFunction = element.toAstFunction()
+    val returnType = astFunction.returnType
+    val returnTypeClass by lazy { returnType.resolvedType().toAstClass() }
```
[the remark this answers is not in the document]
eygraber: It was for cleanliness later in the function too avoid resolving twice. I can rework that.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TargetComponentAccessorGenerator.kt
```diff
@@ -0,0 +1,70 @@
...
+                        val returnTypeCompanion = when {
+                            options.generateCompanionExtensions -> returnTypeClass.companion?.type
+                            else -> null
+                        }
+
+                        val returnTypeName = returnTypeClass.type.toTypeName()
+
+                        addCode(
+                            CodeBlock.of(
+                                "return %T.create($funcParams)",
```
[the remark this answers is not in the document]
eygraber: I found [this](https://github.com/evant/kotlin-inject/issues/371) while testing :smile:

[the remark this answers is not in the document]
evant: Hah good catch!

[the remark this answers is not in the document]
evant: Looks like it should be using '%N' instead of '%L'? https://square.github.io/kotlinpoet/n-for-names/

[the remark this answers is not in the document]
eygraber: Yes `%N` fixes it. I fixed #371 with it as well

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/ProcessTargetComponentAccessor.kt
```diff
@@ -0,0 +1,60 @@
...
+    if (!astFunction.validateReturnType(returnTypeClass, provider)) return true
+
+    process(astFunction, returnTypeClass, codeGenerator, targetComponentAccessorGenerator)
+    true
+}
+
+private fun AstFunction.validateIsExpect(provider: KSAstProvider) = isExpect.also { isValid ->
+    if (!isValid) {
+        provider.error("$name should be an expect fun", this)
+    }
```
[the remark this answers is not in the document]
eygraber: I figured in this case it can't become an expect fun in later rounds anyways, so just report the issue right away. But I can move it to the generator.

eygraber:
Added some documentation for KMP in general (mostly generated by ChatGPT, with some minor human intervention).

eygraber:
@evant anything blocking this from getting merged?

evant:
> Added some documentation for KMP in general (mostly generated by ChatGPT, with some minor human intervention).

I took a quick look at it and it's confusing and unhelpful, I can't accept this in this state. I would go into details but I'm not spending more time one it than you took to generate it.

eygraber:
@evant I updated the documentation to be fully human written :sweat_smile:

evant:
Yeah much better lol. Actually there was a start of this already with https://github.com/evant/kotlin-inject/pull/339, I'm going to make some adjustments to it and merge it in first, then we can combine it with the new annotation instructions here. I'm also working on updating the [greeter sample](https://github.com/evant/kotlin-inject-samples/tree/main/multiplatform/greeter) to better align, albeit with manual expect/actual for now.

evant:
https://github.com/evant/kotlin-inject-samples/pull/19

evant:
Ok merged, would suggest the last section should be replaced with annotation use, also feel free to pull anything else your wrote over if you feel it would be helpful. (I think the first section mentioning the tradoffs about generating for the common source set would be good too)

eygraber:
I'll work on merging my docs in; hope to have it done some time next week.

I also thought of a much better annotation name than TargetComponentAccessor, but I forgot to write it down and I forgot it :see_no_evil: 

While writing the docs it didn't feel like TargetComponentAccessor is really a great name. It might be technically correct, but I don't think it communicates its purpose to users. I'm going to spend some time trying to remember the better name that I though of.

Another thing I thought of was keeping it in a separate kmp runtime artifact, so that users not using kmp don't accidentally use it (it might also help with naming the annotation because it can be more on the nose about being for multiplatform). What do you think?

evant:
> While writing the docs it didn't feel like TargetComponentAccessor is really a great name

Yep, absolutely agree, I'll let you know if I think of anything too. After updating the sample project I do feel like I have a bit more handle on the problem than I did before.

> Another thing I thought of was keeping it in a separate kmp runtime artifact

Hm, yeah it's a hard to know long-term which would be the right decision there, but that's something we can always change later if we need to so I say go for it.

eygraber:
Pushed a bunch of updates. I made some changes to the existing docs. I'll annotate my reasoning there, but I'm fine to revert any of it if you feel that it shouldn't be changed.

README.md
```diff
@@ -35,11 +35,11 @@ val appComponent = AppComponent::class.create()
 val repo = appComponent.repo
 ```
 
-## Download 
+## Gradle 
 
 Using [ksp](https://github.com/google/ksp)
 
-`settings.gradle`
+`settings.gradle.kts`
```
eygraber, line 42: [Kotlin DSL is the default now](https://blog.gradle.org/kotlin-dsl-is-now-the-default-for-new-gradle-builds) so I thought this might be appropriate.

README.md
```diff
@@ -50,36 +50,24 @@ pluginManagement {
...
-`build.gradle`
+`build.gradle.kts`
 
-```groovy
+```kotlin
 plugins {
-    id("org.jetbrains.kotlin.jvm") version "1.9.0"
-    id("com.google.devtools.ksp") version "1.9.0-1.0.13"
+    id("org.jetbrains.kotlin.jvm")
+    id("com.google.devtools.ksp")
```
eygraber, line 58: I thought that using specific versions might give the impression that kotlin-inject will only work with those versions

README.md
```diff
@@ -50,36 +50,24 @@ pluginManagement {
...
-```
-
-### or with KAPT (deprecated)
-
-```groovy
-plugins {
-  id("org.jetbrains.kotlin.jvm") version "1.9.0"
-  id("org.jetbrains.kotlin.kapt") version "1.9.0"
-}
+val kotlinInjectVersion = "<latest kotlin-inject version>"
```
eygraber, line 66: Same reasoning here. It's likely that this will get out of date and cause confusion (although I do have several libraries where I update the version in the README as part of the release process).

README.md
```diff
@@ -50,36 +50,24 @@ pluginManagement {
...
-}
-```
-
-### or with KAPT (deprecated)
-
-```groovy
-plugins {
-  id("org.jetbrains.kotlin.jvm") version "1.9.0"
-  id("org.jetbrains.kotlin.kapt") version "1.9.0"
-}
```
eygraber, line 78: KAPT support was removed (although the latest stable still has it)

docs/multiplatform.md
```diff
@@ -4,82 +4,58 @@ This document covers using kotlin-inject in a multiplatform project. The example
...
-    repositories {
-        mavenCentral()
-        google()
-    }
-}
-```
-
-`build.gradle`
-
-Add Plugin
```
eygraber, line 31: I removed all of this because it is already in the README

docs/multiplatform.md
```diff
@@ -4,82 +4,58 @@ This document covers using kotlin-inject in a multiplatform project. The example
...
-        gradlePluginPortal()
-    }
-}
-
-dependencyResolutionManagement {
-    repositories {
-        mavenCentral()
-        google()
-    }
-}
```
eygraber, line 26: I personally use this in my projects but it's still an unstable API (marked as `@Incubating` in Gradle)

docs/multiplatform.md
```diff
@@ -4,82 +4,58 @@ This document covers using kotlin-inject in a multiplatform project. The example
...
+    add("kspIosX64", "me.tatarka.inject:kotlin-inject-compiler-ksp:$kotlinInjectVersion")
+    add("kspIosArm64", "me.tatarka.inject:kotlin-inject-compiler-ksp:$kotlinInjectVersion")
+    add("kspIosSimulatorArm64", "me.tatarka.inject:kotlin-inject-compiler-ksp:$kotlinInjectVersion")
+    add("kspAndroid", "me.tatarka.inject:kotlin-inject-compiler-ksp:$kotlinInjectVersion")
 }
 ```
 
-Note: You'll have to add the compiler for each of the chipset separately like done above.
-KSP will eventually have better multiplatform support and we'll be able to simply
-have `ksp(libs.kotlinInject.compiler)` :crossed_fingers:
```
eygraber, line 81: In a later commit I linked to the KSP PR that is likely to be the one to change how this works (if it ever changes)

kotlin-inject-runtime/src/commonMain/kotlin/me/tatarka/inject/annotations/Annotations.kt
```diff
@@ -32,4 +32,4 @@ annotation class Assisted
 annotation class Qualifier
 
 @Target(FUNCTION)
-annotation class TargetComponentAccessor
+annotation class KmpComponentCreator
```
eygraber, line 35: I like `KmpComponentProvider` better, but I didn't want to cause any confusion with `@Provides`

README.md
evant: > although I do have several libraries where I update the version in the README as part of the release process

yep that's my standard practice, I hate it when I have to hunt down what the latest version is lol

evant: I've also steered away form `val whateverVersion = ` just because there's like 20 different ways to do this in gradle and I think the plain string is the easiest place to start from a copy & paste prospective.

README.md
evant: Hm, fair enough. It's so you can copy those lines and have it actually work, I tend to update these to whatever that latest is at the time of each release but you are right that it might be giving a false impression.

README.md
evant: yeah, I plan to do a README refresh with the next release, It's the first thing you see when navigating to the project on github so I don't love updating it before a release is actually out.

docs/multiplatform.md
evant: 👍 I had left it in but was considering doing the same.

docs/multiplatform.md
```diff
@@ -4,128 +4,186 @@ This document covers using kotlin-inject in a multiplatform project. The example
...
 ```
 
-Add Runtime dependency in commonMain
+Add the runtime dependency in commonMain
 ```kotlin
 sourceSets {
     commonMain {
         dependencies {
-            implementation("me.tatarka.inject:kotlin-inject-runtime:0.6.3")
+            implementation("me.tatarka.inject:kotlin-inject-kmp-runtime:$kotlinInjectVersion")
```
evant, line 43: while I like to keep the version up-to-date in the README, yeah it would probably be best to leave it out of the other docs because it would be easy to get out of date. Would rather do something like `implementation("me.tatarka.inject:kotlin-inject-kmp-runtime:<latest version>")`
though as it's easier to copy/paste and edit imo.

kotlin-inject-runtime/src/commonMain/kotlin/me/tatarka/inject/annotations/Annotations.kt
evant: Yeah, that's a better name! Curious what you'd think about `@CreateKmpComponent`? Feel like it reads slightly better but open to feedback.

settings.gradle.kts
```diff
@@ -16,6 +16,7 @@ dependencyResolutionManagement {
 rootProject.name = "kotlin-inject"
 
 include(":kotlin-inject-runtime")
+include(":kotlin-inject-kmp-runtime")
```
evant, line 19: I think `kotlin-inject-runtime-kmp` would match the ordering like `kotlin-inject-compiler-ksp` a bit better.

evant (commented):
Made some more specific comments but I think it's worth keeping the README the same for now.

README.md
eygraber: In some projects specifying the versions can cause an error 😅

In general my approach with these things is that the README for libraries don't need to show how to set up the whole project.

README.md
eygraber: Here's how I keep the versions updated - https://github.com/eygraber/uri-kmp/blob/master/.github/workflows/publish_release.yml#L123

I went with the `val` just so it wasn't repeated everywhere, but if it's updated automatically it matters less.

docs/multiplatform.md
eygraber: If you're open to having it kept up to date when releasing that could work for these files as well.

kotlin-inject-runtime/src/commonMain/kotlin/me/tatarka/inject/annotations/Annotations.kt
eygraber: I think both work. It's action vs descriptor, i.e. verb vs noun.

Do you want to make the call?

README.md
eygraber: I reverted the changes to the README

docs/multiplatform.md
eygraber: For now I actually changed it to 0.7.0-SNAPSHOT since it wouldn't actually work on 0.6.3

eygraber:
Addressed feedback, and waiting for decision on [@KmpComponentCreator vs @CreateKmpComponent](https://github.com/evant/kotlin-inject/pull/370/files#r1587015361)

docs/multiplatform.md
evant: Yeah that would make things more clear until we release

evant:
Let's go with CreateKmpComponent then

eygraber:
Done!

eygraber:
I thought about it a bit more and `CreateKmpComponent` is technically a misnomer because the Component isn't "KMP".

Maybe `CreateComponentForKmp` would be more accurate?

evant:
Hm, yeah. I'm starting to second-guess my suggestion too, we are really focusing on the `create` here and not the `Component`. How about `@KmpComponentCreate`? I know it's basically the same as what you originally suggested heh.

eygraber:
Haha that works for me. I'll push that in a couple of minutes.

evant (approved):

PR #374, opened 2024-04-09
Make the commonJs source set depend on commonMain

eygraber:


evant:
Interesting, now I'm wondering if anything was broken before?

eygraber:
I think it was fine before because common could read platform sources. That changed in Kotlin 2.0 but I figured this could be done before that (and it would technically be correct), so why not

PR #376, opened 2024-04-17
Update overridden functions when collecting types

### LINKED ISSUE #375: Implemented function still considered abstract
Imagine this example:
```kotlin
interface Interface1 {
  fun string(): String
}
                    
interface Interface2 : Interface1 {
  override fun string(): String = "abc"
}

@Component 
abstract class MyComponent1 : Interface1, Interface2
```
This fails with:
```
Cannot find an @Inject constructor or provider for: String
```
But if I change the order of the super types to the following, then the code compiles successfully:
```
@Component 
abstract class MyComponent1 : Interface2, Interface1
```

vRallev:
The old mechanism depended on the order types were defined and processed them in order without respecting overridden functions. With the fix overridden functions get updated when collecting types.

Fixes #375

### DIFF
--- a/kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/InheritanceTest.kt
+++ b/kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/InheritanceTest.kt
@@ -1,0 +1,47 @@
+package me.tatarka.inject.test
+
+import me.tatarka.inject.ProjectCompiler
+import me.tatarka.inject.Target
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.api.io.TempDir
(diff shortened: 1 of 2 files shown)


kotlin-inject-compiler/test/src/test/kotlin/me/tatarka/inject/test/InheritanceTest.kt
```diff
@@ -0,0 +1,47 @@
...
+                    interface Interface1 {
+                        fun string(): String
+                    }
+                    
+                    interface Interface2 : Interface1 {
+                        override fun string(): String = "abc"
+                    }
+
+                    @Component 
+                    abstract class MyComponent1 : Interface1, Interface2
```
vRallev, line 39: For this class it used to fail.

evant (approved):

PR #379, opened 2024-04-23
When comparing scope annotations, take their values into account

### LINKED ISSUE #377: Support arguments for scope annotations
For each scope we need to define a new scope annotation, e.g.
```kotlin

@Scope annotation class SingleInAppScope
@Scope annotation class SingleInLoggedInScope
@Scope annotation class SingleInRendererScope
...
```
Rather, I'd like to define a single annotation and use different arguments:
```kotlin
@Scope
annotation class SingleIn(val scope: KClass<*>)
```
Then I only need to define new class references and can reuse the same scope:
```kotlin
@SingleIn(AppScope::class)
@SingleIn(LoggedInScope::class)
...
```
The challenge is that more argument types such as ints, strings and enums must be supported for this.

Note that scope annotations using this mechanism today compile fine, but `@SingleIn(AppScope::class)` would be treated the same as `@SingleIn(LoggedInScope::class)`. The argument is ignored for comparisons. 

The same should be supported for qualifiers #253.

evant:
ex: @NamedScope("one") and @NamedScope("two") are now treated as distinct scopes.

Fixes #377

### DIFF
--- a/integration-tests/common/src/main/kotlin/me/tatarka/inject/test/Scope.kt
+++ b/integration-tests/common/src/main/kotlin/me/tatarka/inject/test/Scope.kt
@@ -236,3 +236,25 @@ abstract class SuperclassScopeFooComponent : BaseSuperclassScopeFooComponent {
 
     companion object
 }
+
+@Scope
+annotation class ScopeWithArg(val value: String)
+
+@Inject
+@ScopeWithArg("parent")
+class ScopedParentFoo
+
+@Inject
+@ScopeWithArg("child")
+class ScopedChildFoo
+
+@Component
+@ScopeWithArg("parent")
+abstract class ParentScopeWithArgComponent
+
+@Component
+@ScopeWithArg("child")
+abstract class ChildScopeWithArgComponent(@Component val parent: ParentScopeWithArgComponent) {
+    abstract val parentFoo: ScopedParentFoo
+    abstract val childFoo: ScopedChildFoo
+}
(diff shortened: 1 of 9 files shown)


integration-tests/common/src/main/kotlin/me/tatarka/inject/test/Scope.kt
```diff
@@ -236,3 +236,25 @@ abstract class SuperclassScopeFooComponent : BaseSuperclassScopeFooComponent {
...
+@Inject
+@ScopeWithArg("parent")
+class ScopedParentFoo
+
+@Inject
+@ScopeWithArg("child")
+class ScopedChildFoo
+
+@Component
+@ScopeWithArg("parent")
```
vRallev, line 252: Curious, `KClass` arguments are supported as well, right? I didn't see how you implement the `equals()` check.

evant: They should be yes, it's using the same logic that was already used for qualifiers

evant: If you are curious https://github.com/evant/kotlin-inject/blob/094d8ce2c1a70a1b29fc08be073052fcf3a76e4a/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSUtil.kt#L60

vRallev: Thanks!

PR #385, opened 2024-05-02
Update the generated filename for KmpComponentCreate functions

eygraber:
I missed the `Target...Accessor` part in the previous PR.

I also simplified the name, because what I had before was pretty gnarly since it was trying to avoid `FileAlreadyExistsException` in case there are multiple `@KmpComponentCreate` functions for the same `Component`.

With this new approach the following would cause a compile time error because the processor would generate two files named `KmpComponentCreateMyComponent.kt`:

```kotlin
@Component
abstract class MyComponent

@KmpComponentCreate
expect fun createKmp(): MyComponent

@KmpComponentCreate
expect fun MyComponent.Companion.createKmp(): MyComponent
```

Using the approach from before there would be no `FileAlreadyExistsException`, but the generated file names would be `KmpComponentCreateMyComponentCreateKmp.kt` and `KmpComponentCreateMyComponentMyComponentCompanionCreateKmp.kt` respectively. That works, and is safest, so the question is how ugly is too ugly for generated file names?

Another approach would be to defer writing the files until the final round, and then group the functions by `Component` type, and generate one file containing all of the functions for that `Component`.

### DIFF
--- a/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/ProcessKmpComponentCreate.kt
+++ b/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/ProcessKmpComponentCreate.kt
@@ -12,8 +12,7 @@ import me.tatarka.kotlin.ast.KSAstProvider
 internal fun processKmpComponentCreate(
     element: KSFunctionDeclaration,
     provider: KSAstProvider,
-    codeGenerator: CodeGenerator,
-    kmpComponentCreateGenerator: KmpComponentCreateGenerator,
+    kmpComponentCreateFunctionsByComponentType: MutableMap<AstClass, MutableList<AstFunction>>
 ): Boolean = with(provider) {
     val astFunction = element.toAstFunction()
     val returnType = astFunction.returnType
@@ -27,8 +26,20 @@ internal fun processKmpComponentCreate(
     val returnTypeClass = returnType.resolvedType().toAstClass()
     if (!astFunction.validateReturnType(returnTypeClass, provider)) return true
 
-    process(astFunction, returnTypeClass, codeGenerator, kmpComponentCreateGenerator)
+    kmpComponentCreateFunctionsByComponentType.getOrPut(returnTypeClass, ::ArrayList).add(astFunction)
+
     true
+}
+
+internal fun generateKmpComponentCreateFiles(
+    codeGenerator: CodeGenerator,
+    generator: KmpComponentCreateGenerator,
+    kmpComponentCreateFunctionsByComponentType: Map<AstClass, List<AstFunction>>
+) {
+    kmpComponentCreateFunctionsByComponentType.forEach { (componentType, kmpComponentCreateFunctions) ->
+        val file = generator.generate(componentType, kmpComponentCreateFunctions)
+        file.writeTo(codeGenerator, aggregating = false)
+    }
 }
 
 private fun AstFunction.validateIsExpect(provider: KSAstProvider) = isExpect.also { isValid ->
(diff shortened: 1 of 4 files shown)


evant:
Another option would be to group by source file, I've used that successively with ksp before though I don't think it's currently doing that for components in this project. ksp's incremental compilation only works on the file-level too anyway so there shouldn't be an issues if one function in the file changes, though I don't know if there's plans to change that.

eygraber:
I implemented grouping the `KmpComponentCreate` functions into a single file by `Component` type and it was pretty simple. I'll push that in a separate commit now.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/ProcessKmpComponentCreate.kt
```diff
@@ -27,10 +26,22 @@ internal fun processKmpComponentCreate(
...
 }
 
+internal fun generateKmpComponentCreateFiles(
+    codeGenerator: CodeGenerator,
+    generator: KmpComponentCreateGenerator,
+    kmpComponentCreateFunctionsByComponentType: Map<AstClass, List<AstFunction>>
+) {
+    kmpComponentCreateFunctionsByComponentType.forEach { (componentType, kmpComponentCreateFunctions) ->
+        val file = generator.generate(componentType, kmpComponentCreateFunctions)
+        file.writeTo(codeGenerator, aggregating = false)
```
evant, line 41: This would need to be aggregating now right? Since there could be multiple create functions for the same component in different source files. My suggestion to group by file would allow it to be kept as non-aggregating, though I'm not sure there's a meaningful different in performance.

eygraber: Good catch, yes this would need to be aggregating. I could switch to grouping by file, but I think that loses some of the expressiveness of the file name. In this case I think aggregation would be ammortized away since the inject files are aggregating anyways, and they'd likely be closely related. What do you think?

evant: >  think that loses some of the expressiveness of the file name
I'm not sure what you mean by that?
> since the inject files are aggregating anyways
component generation is not aggregating, since there's a file generated for each component, unless you meant something else?

eygraber: > > think that loses some of the expressiveness of the file name

> I'm not sure what you mean by that?

For example, I have a file called `SearchPortal.kt` that contains a `@Component abstract class SearchComponent` and `@KmpComponentCreate expect fun SearchComponent.Companion.createKmp`.

The generated file for the `Component` is `InjectSearchComponent.kt` which lets me know that this is the generated file related to `SearchComponent`.

With my current changes, the generated file for `KmpComponentCreate` would be `KmpComponentCreateSearchComponent.kt` which lets me know that this is the generated file related to `SearchComponent`.

If we switch it to be source file based the generated file would be `KmpComponentCreateSearchPortal.kt` which doesn't necessarily indicate that it is related to `SearchComponent` (the source file might be entirely unrelated, there might be multiple `Components` in the file, etc...).

> > since the inject files are aggregating anyways

> component generation is not aggregating, since there's a file generated for each component, unless you meant something else?

Doesn't [this](https://github.com/evant/kotlin-inject/blob/main/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/ProcessInject.kt#L79) indicate that it is aggregating, or am I confusing something?

evant: 1. Yeah I guess that's true, I don't put that much weight in them since it's not that all that likely you'd be looking at the generated filenames directly but not against making them better.
2. 🤦‍♀️ should've looked before saying that, I don't think it was originally but there was probably a reason to changing to that. "aggregation would be ammortized away" is probably true then.

eygraber: OK, so is this good to merge? I pushed a change setting `aggregating = true`.

evant: yeah

evant (approved):

PR #391, opened 2024-06-08
fix generated code for scoped lazy circular dependency (#313)

### LINKED ISSUE #313: Combination of Lazy, circular dependencies, and Scopes results in invalid generated code
A scoped component that provides a scoped class (Foo) with a Lazy and real dependency on another class (Bar), which in turn has a Lazy dependency on the first class, results in invalid generated code.
```kotlin

@Scope
annotation class CustomScope

@CustomScope
@Component
abstract class BaseComponent {
    abstract val foo: Foo
    abstract val bar: Bar
}

@CustomScope
@Inject
class Foo(
    val lazy: Lazy<Bar>,
    val real: Bar,
)

@Inject
class Bar(
    val foo: Lazy<Foo>
)
```

The resulting generated override for Foo is invalid:
```kotlin
import kotlin.reflect.KClass
import me.tatarka.inject.`internal`.LazyMap
import me.tatarka.inject.`internal`.ScopedComponent

public fun KClass<BaseComponent>.create(): BaseComponent = InjectBaseComponent()

public class InjectBaseComponent() : BaseComponent(), ScopedComponent {
  override val _scoped: LazyMap = LazyMap()

  override val foo: Foo
    get() = run<Foo> {
      lateinit var foo: Foo
      foo.also {
        foo = it
      }
    }

  override val bar: Bar
    get() = Bar(
      foo = lazy {
        foo
      }
    )
}
```

sergeshustoff:
* added test for case with scoped circular lazy dependency #313
* fix for the test case by not wrapping TypeResult.Scoped with TypeResult.LateInit because wrapping is already done for TypeResult.Scoped.result if it was needed

Fixes #313

### DIFF
--- a/integration-tests/common-test/src/test/kotlin/me/tatarka/inject/test/RecursiveTest.kt
+++ b/integration-tests/common-test/src/test/kotlin/me/tatarka/inject/test/RecursiveTest.kt
@@ -24,6 +24,14 @@ class RecursiveTest {
     }
 
     @Test
+    fun generates_a_component_that_provides_a_lazy_scoped_dea_recursively() {
+        val component: LazyScopedCycleComponent = LazyScopedCycleComponent::class.create()
+        val foo = component.foo
+
+        assertThat(foo).isSameInstanceAs(foo.lazy.value.foo.value)
+    }
+
+    @Test
     fun generates_a_component_that_provides_a_function_dea_recursively() {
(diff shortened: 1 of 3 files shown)


evant (approved):
Thanks! The whole logic around this really needs an overhaul but this is a simple solution for now

PR #392, opened 2024-06-11
Fix `KtInvalidLifetimeOwnerAccessException` with KSP2

vRallev:
`InjectProcessor` remembers deferred symbols across rounds. KSP (KSP 2 in particular) forbids to cache symbols across rounds and verifies the lifecycle. The fix is trying to get new symbols of the deferred classes each round.

For more details: https://github.com/google/ksp/issues/1854

### DIFF
--- a/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
+++ b/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
@@ -37,7 +37,15 @@ class InjectProcessor(
         injectGenerator = InjectGenerator(provider, options)
         kmpComponentCreateGenerator = KmpComponentCreateGenerator(provider, options)
 
-        val previousDeferredClasses = deferredClasses
+        val previousDeferredClasses = deferredClasses.map { original ->
+            // This attempts to reload / refresh the symbol of the deferred classes. KSP forbids to cache symbols
+            // across rounds, which is what deferredClasses is doing. As an alternative,
+            // getSymbolsWithClassAnnotation() could look at all files and not only newly added files in this round
+            // and not remember deferred symbols.
+            //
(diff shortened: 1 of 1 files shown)


evant:
I think we need to do the same thing with deferredFunctions too right?

evant:
Might be a little cleaner to only hold on to the qualified names between rounds, that way it _has_ to be looked up again instead of having to remember to.

evant:
We really should have some more rounds tests to verify things like this, not going to hold up this pr on that though

vRallev:
> I think we need to do the same thing with deferredFunctions too right?

Yes, but I'm not sure what these function cover or how to "refresh" the symbols. Are these only top-level functions? I also couldn't test this whereas I had reproducible steps for classes and wanted to start with that.

> Might be a little cleaner to only hold on to the qualified names between rounds, that way it has to be looked up again instead of having to remember to.

I thought about that, but decided against it in the end. Reason is that APIs like `qualifiedName` are nullable and I'm falling back to the original symbol in case we need to cover anonymous classes or other rare cases. Not all APIs on a symbol throw this exception.  

I really believe theres a lack of API support in KSP.

evant:
> Are these only top-level functions?

Yes, they were added to support [KmpComponentCreate](https://github.com/evant/kotlin-inject/blob/main/docs/multiplatform.md#shared-source-sets)

evant:
> in case we need to cover anonymous classes or other rare cases

I only expect well ever be scanning classes & top-level functions which are always supported right? Falling back to the original symbol that is 'unsupported' sounds like a recipe for some rare hard-to-debug issues to me.

evant:
https://github.com/evant/kotlin-inject/pull/393 is what I mean. You don't need to look up the symbols again except in the finish case which should hopefully be less common as it implies something was never resolved.

vRallev:
I left comments. Let's continue the discussion there and then merge one or the other fix.

evant:
Superseded by https://github.com/evant/kotlin-inject/pull/393

PR #393, opened 2024-06-11
Don't reuse symbols between rounds

evant:
Reusing symbols between rounds is not supported and throws an error on ksp2. Unfortunately this means we need to use `getSymbolsWithAnnotation` instead of our custom functions as they are currently the only way to access updated deferred symbols. We still keep a list of the symbol names so that we can run a final processing pass on the last round.

See https://github.com/google/ksp/issues/1854

### DIFF
--- a/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
+++ b/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
@@ -45,14 +45,26 @@ class InjectProcessor(
         val deferredClasses = componentSymbols.filterNot { element ->
             processInject(element, provider, codeGenerator, injectGenerator)
         }.toList()
-        deferredClassNames = deferredClasses.mapNotNull { it.qualifiedName }
+        deferredClassNames = deferredClasses.mapNotNull {
+            val name = it.qualifiedName
+            if (name == null) {
+                logger.warn("Unable to defer symbol: ${it.simpleName.asString()}, no qualified name", it)
+            }
+            name
+        }
 
         val kmpComponentCreateSymbols = resolver.getSymbolsWithAnnotation(KMP_COMPONENT_CREATE.canonicalName)
             .filterIsInstance<KSFunctionDeclaration>()
         val deferredFunctions = kmpComponentCreateSymbols.filterNot { element ->
             processKmpComponentCreate(element, provider, kmpComponentCreateFunctionsByComponentType)
         }.toList()
-        deferredFunctionNames = deferredFunctions.mapNotNull { it.qualifiedName }
+        deferredFunctionNames = deferredFunctions.mapNotNull {
+            val name = it.qualifiedName
+            if (name == null) {
+                logger.warn("Unable to defer symbol: ${it.simpleName.asString()}, no qualified name", it)
+            }
+            name
+        }
 
         return deferredClasses + deferredFunctions
     }
@@ -64,7 +76,7 @@ class InjectProcessor(
             for (name in deferredClassNames) {
                 val element = resolver.getClassDeclarationByName(name)
                 if (element == null) {
-                    logger.error("Failed to resolve: $name")
+                    logger.error("Failed to resolve: ${name.asString()}")
                     continue
                 }
                 processInject(
@@ -82,7 +94,7 @@ class InjectProcessor(
                     includeTopLevel = true
                 ).firstOrNull()
                 if (element == null) {
-                    logger.error("Failed to resolve: $name")
+                    logger.error("Failed to resolve: ${name.asString()}")
                     continue
                 }
                 processKmpComponentCreate(element, provider, kmpComponentCreateFunctionsByComponentType)
--- a/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
+++ b/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
@@ -9,6 +9,7 @@ import com.google.devtools.ksp.processing.SymbolProcessorProvider
 import com.google.devtools.ksp.symbol.KSAnnotated
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSFunctionDeclaration
+import com.google.devtools.ksp.symbol.KSName
 import me.tatarka.inject.compiler.COMPONENT
 import me.tatarka.inject.compiler.InjectGenerator
 import me.tatarka.inject.compiler.KMP_COMPONENT_CREATE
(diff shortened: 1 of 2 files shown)


kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
```diff
@@ -27,63 +28,69 @@ class InjectProcessor(
...
-        val previousDeferredClasses = deferredClasses
-        val previousDeferredFunctions = deferredFunctions
-
-        val componentSymbols = previousDeferredClasses + resolver.getSymbolsWithClassAnnotation(
-            packageName = COMPONENT.packageName,
-            simpleName = COMPONENT.simpleName
-        )
-        deferredClasses = componentSymbols.filterNot { element ->
+        val componentSymbols =
+            resolver.getSymbolsWithAnnotation(COMPONENT.canonicalName).mapNotNull { it as? KSClassDeclaration }
```
vRallev, line 44: Could use `filterIsInstance` instead.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
```diff
@@ -27,63 +28,69 @@ class InjectProcessor(
...
+        val deferredClasses = componentSymbols.filterNot { element ->
             processInject(element, provider, codeGenerator, injectGenerator)
-        }
+        }.toList()
+        deferredClassNames = deferredClasses.mapNotNull { it.qualifiedName }
```
vRallev, line 48: If we know that the qualified name must not be null, then I wouldn't use `mapNotNull` and crash for a null value instead. Silently dropping a deferred class sounds scary to me.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
```diff
@@ -27,63 +28,69 @@ class InjectProcessor(
...
     override fun finish() {
-        // Last round, generate as much as we can, reporting errors for types that still can't be resolved.
-        for (element in deferredClasses) {
-            processInject(
-                element,
-                provider,
+        try {
+            // Last round, generate as much as we can, reporting errors for types that still can't be resolved.
+            val resolver = lastResolver ?: return
+            for (element in deferredClassNames.mapNotNull { resolver.getClassDeclarationByName(it) }) {
```
vRallev, line 64: Same here, I would not silently drop classes.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
```diff
@@ -27,63 +28,69 @@ class InjectProcessor(
...
     override fun process(resolver: Resolver): List<KSAnnotated> {
+        lastResolver = resolver
         provider = KSAstProvider(resolver, logger)
         injectGenerator = InjectGenerator(provider, options)
         kmpComponentCreateGenerator = KmpComponentCreateGenerator(provider, options)
 
-        val previousDeferredClasses = deferredClasses
-        val previousDeferredFunctions = deferredFunctions
-
-        val componentSymbols = previousDeferredClasses + resolver.getSymbolsWithClassAnnotation(
```
vRallev, line 43: That's a behavior change. Shouldn't you still try to generate code for deferred classes from previous rounds and not wait until `finish()`? This would allow other symbol processor to react to the generated code.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
evant: I originally had that but then decided to drop was ok because this only runs in the case where the symbol gets deferred past the last round (when a dependent symbol is _never_ resolved). This implies other errors in the project so I think it's acceptable to be dropped here if that ever happens.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
evant: As per https://github.com/google/ksp/issues/1854#issuecomment-2161292017 `getSymbolsWithAnnotation` adds i the deferred symbols so we don't have to manually do it ourselves.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
vRallev: > This implies other errors in the project

Please highlight this by throwing an exception. Silently dropping makes it much harder to diagnose issues.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
vRallev: Wouldn't this mean you generate the same component now twice? This would result in exceptions. E.g. round 1 symbol is deferred, round 2 you generate the code in `process()`, `finish()` gets called and you generate the class one more time.

vRallev: Oh never mind, it's 
```
deferredClassNames = deferredClasses.mapNotNull { it.qualifiedName }
```
And not
```
deferredClassNames += deferredClasses.mapNotNull { it.qualifiedName }
```

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
```diff
@@ -27,63 +28,69 @@ class InjectProcessor(
...
-        }
-
-        generateKmpComponentCreateFiles(
-            codeGenerator,
-            kmpComponentCreateGenerator,
-            kmpComponentCreateFunctionsByComponentType
-        )
-        kmpComponentCreateFunctionsByComponentType.clear()
-
-        deferredFunctions = mutableListOf()
```
vRallev, line 86: I'll try the changes with my breaking tests later tonight and will report back if that works.

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
evant: Fair enough, I did tell you the same thing 😅

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
evant: done

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
evant: done

vRallev: I still see `mapNotNull`. Any chance this could be changed, too?

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
vRallev: Seems to work great!

kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
evant: ooh, if qualifiedName is null? Yeah, sure

evant: done, chose to make it a warning since I don't really understand the circumstanced when it might be null and don't want to prevent cases where it is resolvable in a future round.

vRallev (approved):

PR #406, opened 2024-06-22
Add clarifying message when provider for a dependency with platform nullability wasn't found

### LINKED ISSUE #401: Can't inject provided dependency into java class
Problem:
when there is a java class in dependency graph we have to create `@Provides` function in component in order to inject it - having `@Inject` annotation on that's java class' constructor doesn't work.

I'm not sure if it's ksp limitation or not, will need to research that...

sergeshustoff:
Closes #401


evant:
hm, I wonder if we should go-ahead and treat injections into java code non-nullable by default.

sergeshustoff:
You're right, I'll try that instead

PR #410, opened 2024-06-29
Consider platform nullability non-null in params

### LINKED ISSUE #401: Can't inject provided dependency into java class

sergeshustoff:
Allows injecting provided dependencies into java constructor even if parameter nullability is not specified in java

Fixes #401

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,7 +74,7 @@ jobs:
   build-macos:
     macos:
       xcode: "14.1.0"
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.medium.gen1
     <<: *defaults
     steps:
       - checkout
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
@@ -91,7 +91,11 @@ class TypeResultResolver(private val provider: AstProvider, private val options:
         var assistedFailed = false
         val args = context.args.toMutableList()
         for (param in params) {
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ jobs:
   build-macos:
     macos:
       xcode: "14.1.0"
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.medium.gen1
```
evant, line 77: :+1: had planned to do this but they new resources weren't available at the time

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
```diff
@@ -190,6 +190,15 @@ class TypeResultResolver(private val provider: AstProvider, private val options:
         return paramsWithName
     }
 
+    /**
+     * Converts platform nullability to non-null. It's only viable for parameters in java code, never for provided types
+     */
+    private fun withPlatformNullabilityAsNotNull(type: AstType) = if (type.isPlatform()) {
```
evant, line 196: minor: a function seems overkill since this is only used in the one place (well there's also resolveParamsLegacy but that errors out now and should be removed for the next release anyway)

evant (approved):

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultResolver.kt
sergeshustoff: ok, removed function

Issue #409, opened 2024-06-26
Support qualifieres in multibinds map keys and set elements. 


evant:
As part of https://github.com/evant/kotlin-inject/issues/253 we expect to support placing a qualifier on a typealias as a replacement to using the typealias is a qualifier directly, ex: `typealias MyString = @Named("name") String`. Right now this causes and issue with multibinds as you can't use in the Map or Set type.

```
abstract val myStringMap: Map<MyString, String>
```

runs into

> e: [ksp] Qualifier: @Named(value=MyString) can only be applied to the outer type

Since we treat these Map and Set here specially we can support the nested qualifier in these cases

sergeshustoff:
Should it be applicable to both key and value?

evant:
Good question, I don't actually know that the current behavior is for multimaps with the same key type but different value types.

evant:
update: same key but different value type is treated as distinct, I can see usecases for this so it's best to keep, which means that yes this should be applicable to both key and value

sergeshustoff:
I've researched a little and I see a few problems:
1. It's kinda weird that generic type qualifiers are conditionally legal and it might be difficult to see if it should be legal or not. Even if it's simplified to only illegal for anything that `@Provides`, it complicates api

        @Component
        abstract class SetComponent {
            abstract val conditionallyLegal: Set<@Named("foo") String>
        
            @Provides
            @IntoSet
            fun makesFirstOneLegal(): @Named("foo") String = "1"
            
            @Provides
            fun butThisIsIllegalAndMakesFirstOneIllegal(): Set<@Named("foo") String> = setOf("1")
        }

2.  `abstract val legal1: Set<@Named("foo1") String>` and  `abstract val legal2: Set<@Named("foo2") String>` would have equal TypeKey (but not SetKey). 

To avoid both problems we would need to include generic's qualifiers (recursively?) into TypeKey comparison and that leads us to maybe supporting type qualifiers for generics in general rather than exclusively for `@IntoSet` and `@IntoMap` cases. Wdyt?

evant:
> To avoid both problems we would need to include generic's qualifiers (recursively?)

I started to take a stab at this and reached a similar conclusion.

sergeshustoff:
What do you think of separating type qualifiers from member qualifiers? It would make folowing 2 things work differently:

    @Named("foo") val someting: String
    val somethingElse: @Named("foo") String

[...]

PR #411, opened 2024-07-02
Fix `NullPointerException` when resolving types

vRallev:
KSP indicates that a type can for a type argument can be null, but kotlin-inject ignored that. This caused crashes in our case:
```
java.lang.NullPointerException
	at me.tatarka.kotlin.ast.KSAstType.getArguments(KSAst.kt:469)
	at me.tatarka.inject.compiler.InjectGeneratorKt.qualifier$checkTypeArgs(InjectGenerator.kt:256)
	at me.tatarka.inject.compiler.InjectGeneratorKt.qualifier$checkTypeArgs(InjectGenerator.kt:261)
	at me.tatarka.inject.compiler.InjectGeneratorKt.qualifier$checkTypeArgs(InjectGenerator.kt:261)
	at me.tatarka.inject.compiler.InjectGeneratorKt.qualifier$qualifier(InjectGenerator.kt:278)
	at me.tatarka.inject.compiler.InjectGeneratorKt.qualifier(InjectGenerator.kt:283)
	at me.tatarka.inject.compiler.TypeCollector.collectTypeInfo(TypeCollector.kt:391)
	at me.tatarka.inject.compiler.TypeCollector.collect(TypeCollector.kt:16)
	at me.tatarka.inject.compiler.TypeCollector.collect$default(TypeCollector.kt:15)
	at me.tatarka.inject.compiler.InjectGenerator.collectTypes(InjectGenerator.kt:177)
	at me.tatarka.inject.compiler.InjectGenerator.generateInjectComponent(InjectGenerator.kt:85)
	at me.tatarka.inject.compiler.InjectGenerator.generate(InjectGenerator.kt:68)
	at me.tatarka.inject.compiler.ksp.ProcessInjectKt.process(ProcessInject.kt:78)
	at me.tatarka.inject.compiler.ksp.ProcessInjectKt.processInject(ProcessInject.kt:28)
	at me.tatarka.inject.compiler.ksp.ProcessInjectKt.processInject$default(ProcessInject.kt:19)
	at me.tatarka.inject.compiler.ksp.InjectProcessor.process$lambda$0(InjectProcessor.kt:46)
	at kotlin.sequences.FilteringSequence$iterator$1.calcNext(Sequences.kt:171)
	at kotlin.sequences.FilteringSequence$iterator$1.hasNext(Sequences.kt:194)
	at kotlin.sequences.SequencesKt___SequencesKt.toList(_Sequences.kt:813)
	at me.tatarka.inject.compiler.ksp.InjectProcessor.process(InjectProcessor.kt:47)
```

I tried reproducing this issue, but I believe it has to do with certain types coming from pre-compiled artifacts. The crash occurred for us with:
```kotlin
Pair<KClass<out BaseModel>, () -> Renderer<*>>
```
Where for `Renderer<*>` the argument is resolved as star variance without any further type.

### DIFF
--- a/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
+++ b/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
@@ -466,7 +466,10 @@ private class KSAstType private constructor(
 
     override val arguments: List<AstType>
         get() = type.arguments.map {
-            KSAstType(resolver, it.type!!)
+            val argumentType = checkNotNull(it.type) {
+                "Couldn't resolve type for $it with variance ${it.variance} from parent type $type"
+            }
+            KSAstType(resolver, argumentType)
         }
 
     override val isError: Boolean
--- a/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
+++ b/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
@@ -203,6 +203,10 @@ abstract class AstType : AstElement(), AstAnnotated {
 
     abstract val simpleName: String
 
+    /**
+     * Returns a list of type arguments. This method may throw an [IllegalStateException] if the type
+     * cannot be resolved.
+     */
     abstract val arguments: List<AstType>
 
     abstract val isError: Boolean
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
@@ -253,7 +253,16 @@ fun <E> qualifier(
 ): AstAnnotation? where E : AstElement, E : AstAnnotated {
     // check for qualifiers incorrectly applied to type arguments
     fun checkTypeArgs(packageName: String, simpleName: String, type: AstType) {
-        for (typeArg in type.arguments) {
+        @Suppress("SwallowedException")
+        val arguments = try {
+            type.arguments
+        } catch (e: IllegalStateException) {
+            // We do a deep analysis of all types, but not all types are necessarily resolvable, e.g.
+            // this may happen with star projections. In this case stop checking type arguments.
+            return
+        }
+
+        for (typeArg in arguments) {
             val argQualifier = typeArg.annotationAnnotatedWith(packageName, simpleName)
             if (argQualifier != null) {
                 provider.error("Qualifier: $argQualifier can only be applied to the outer type", typeArg)


ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
```diff
@@ -465,9 +465,14 @@ private class KSAstType private constructor(
...
 
     override val arguments: List<AstType>
-        get() = type.arguments.map {
-            KSAstType(resolver, it.type!!)
-        }
+        get() = type.arguments
+            .mapNotNull {
+                // The type can be null for STAR use-site variance. In some cases KSP then cannot resolve the type.
+                it.type
+            }
```
evant, line 472: I worry that skipping over args would lead to invalid code-gen, maybe we should bail out in this case? Depends on where these arguments are being used I guess.

vRallev: I thought about other options, e.g. making the `type` nullable, but this would cause a lot of cascading changes in `KSAstType`. I thought about implementing `AstType` for star projections, but even then I'd need to throw an exception for certain APIs.

vRallev: I know this is nice to have, but I wonder of kotlin-inject has to do this:
```
        for (typeArg in type.arguments) {
            val argQualifier = typeArg.annotationAnnotatedWith(packageName, simpleName)
            if (argQualifier != null) {
                provider.error("Qualifier: $argQualifier can only be applied to the outer type", typeArg)
            }
            checkTypeArgs(packageName, simpleName, typeArg)
        }
```
Without this loop the problem is gone, too.

evant: The important thing is I'd want to it crash in other places where we'd be handling this incorrectly instead of skipping over the argument and generating invalid code. Did you mean you tried to changing it to `override val arguments: List<AstType?>`? If you don't want to go an inspect every single usage now then putting `!!` should be ok, it'll crash at that site in the future and we can take it from there. Could even add something like

```
fun AstType.expectArguments(): List<AstType>
```
to make it more clear that that is what we are doing.

> Right now it's crashing when checking for qualifiers.

For that specific case, yeah I think skipping it would be acceptable. That code will need to be refactored anyway for https://github.com/evant/kotlin-inject/issues/409 fyi though I guess that's neither here nor there.

evant: > Without this loop the problem is gone, too.

The check was added to punt on more complex qualifier handling, right now only a single qualifier on a type is supported.

vRallev: New revision. Let me know what you think. I kept the crashing behavior for now and handle the exception in this one particular case.

ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
```diff
@@ -466,7 +466,10 @@ private class KSAstType private constructor(
 
     override val arguments: List<AstType>
         get() = type.arguments.map {
-            KSAstType(resolver, it.type!!)
+            val argumentType = checkNotNull(it.type) {
+                "Couldn't resolve type for $it with variance ${it.variance} from parent type $type"
+            }
```
evant, line 471: 👍 on better error message

evant (approved):
Good enough for now, may be worth filing an issue about star projections since this may come up again in the future.

vRallev:
Fixed Detekt

vRallev:
> Good enough for now, may be worth filing an issue about star projections since this may come up again in the future.

I filed #412

PR #413, opened 2024-07-06
Assisted factory interface

### LINKED ISSUE #243: Allow use of `fun interface` for assisted injection

### LINKED ISSUE #252: Support factory interfaces for function injection

sergeshustoff:
- Added assisted factory interface support similar to dagger
- Regular interfaces are supported alongside with functional interfaces for easier migration from dagger
- Function injection is supported with optional parameter in `@AssistedFactory` annotation instead of name matching

Closes #243
Closes #252

### DIFF
--- a/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
+++ b/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
@@ -43,6 +43,8 @@ interface AstAnnotated {
     fun annotationsAnnotatedWith(packageName: String, simpleName: String): Sequence<AstAnnotation>
 }
 
+fun AstAnnotated.hasAnnotation(annotation: ClassName) = hasAnnotation(annotation.packageName, annotation.simpleName)
+
 fun AstAnnotated.annotationAnnotatedWith(packageName: String, simpleName: String): AstAnnotation? =
     annotationsAnnotatedWith(packageName, simpleName).firstOrNull()
 
@@ -247,6 +249,8 @@ abstract class AstAnnotation : AstElement() {
     abstract val type: AstType
 
     abstract fun toAnnotationSpec(): AnnotationSpec
+
+    abstract fun argument(name: String): Any?
 }
 
 abstract class AstParam : AstElement(), AstAnnotated {
--- a/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
+++ b/ast/core/src/main/kotlin/me/tatarka/kotlin/ast/Ast.kt
@@ -149,6 +149,8 @@ abstract class AstFunction : AstMember() {
 
     abstract val packageName: String
 
+    abstract val isTopLevel: Boolean
+
     abstract override fun equals(other: Any?): Boolean
 
     abstract override fun hashCode(): Int
--- a/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
+++ b/ast/ksp/src/main/kotlin/me/tatarka/kotlin/ast/KSAst.kt
@@ -12,6 +12,7 @@ import com.google.devtools.ksp.processing.KSPLogger
 import com.google.devtools.ksp.processing.Resolver
 import com.google.devtools.ksp.symbol.ClassKind
 import com.google.devtools.ksp.symbol.FileLocation
+import com.google.devtools.ksp.symbol.FunctionKind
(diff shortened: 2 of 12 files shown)


sergeshustoff:
TODO:
- [ ] make sure injected function is global or belongs to global object
- [ ] add tests
- [ ] support `@AssistedInject` dagger annotation when enableJavaxAnnotations is on
- [ ] fix detekt issues

evant:
> support @AssistedInject and @Assisted dagger annotations when enableJavaxAnnotations is on

That's an interesting idea, but we don't currently have plans to implement 'dagger compatibility'. The javax annotation features is just for annotations specified by [JSR-330](https://github.com/javax-inject/javax-inject). Would recommend leaving that part out of this pr and open an new issue for it instead.

sergeshustoff:
Ok, I'll remove dagger's assisted factory annotation too. What do you think of non-sam interfaces for assisted factories though? Without an easy migration from dagger in mind it might not make much sense to to use regular interfaces instead of fun interfaces. But on the other hand it might be turned into something entirely different if multiple functions are allowed

evant:
fun interfaces don't allow default args which was one of the motivations for this feature, so yes I think being loser on what we allow is reasonable.

sergeshustoff:
Ready for review. I don't quite like how it turned out with findAssistedFactoryFunction and findAssistedFactoryInjectFunction functions, and how potential name shadowing is handled... But for now I have no idea how to improve that

sergeshustoff:
Decided to go with a single `@AssistedFactory(injectFunction = "optional.function")` and a single function allowed in assisted factory. With multiple functions the injectFunciton name would have to go to a separate annotation. Plus the concern of assisted factory turning into a weird subcomponent.

Potentially this annotation can be later applied to types to have a unified and explicit approach to assisted factories with lambdas too... Though I'm not sure if it's needed - the only case when it helps is the unlikely scenario when some lambdas are provided in the component and it becomes difficult to distinguish them from assisted factory lambdas

sergeshustoff:
@evant is there something we need to do to merge this? To me a few files look a bit bloated, but I wouldn't want to mess up the history by extracting stuff out of them, at least not in this PR

evant:
Based on the responses to https://github.com/evant/kotlin-inject/issues/243#issuecomment-2344994839 I think we are good to move forward on this.

evant (approved):

sergeshustoff:
@Monabr I don't think I have write access to do that

evant:
has merge conflicts, @sergeshustoff can fix or I will eventually when I get a chance

evant:
Screenshot 2024-10-08 at 2 40 19 PM

evant:
@Monabr can you please calm down? This isn't urgent. It will get merged

sergeshustoff:
Weird, GitHub shows me that there are no conflicts. Though maybe my merging main into the branch prevents rebase? I guess I could rebase old changes to latest main, cerrypick newer changes on top of that and open s new pr?

PR #414, opened 2024-07-10
Fix `FileAlreadyExistsException`

vRallev:
Classes keep being processed multiple times and throw a `FileAlreadyExistsException` in multi round scenarios. The function `getSymbolsWithAnnotation()` returns deferred symbols twice for some reason in the next round.

The fix is only applied for classes and not functions yet, because there the issue didn't occur. More care should also be taken for functions, because the qualified name alone is not an indicator to distinguish functions as functions be overloaded and have different parameters.

### DIFF
--- a/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
+++ b/kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
@@ -42,9 +42,19 @@ class InjectProcessor(
 
         val componentSymbols =
             resolver.getSymbolsWithAnnotation(COMPONENT.canonicalName).filterIsInstance<KSClassDeclaration>()
-        val deferredClasses = componentSymbols.filterNot { element ->
-            processInject(element, provider, codeGenerator, injectGenerator)
-        }.toList()
+        val deferredClasses = componentSymbols
+            .distinctBy {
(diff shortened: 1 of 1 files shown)


kotlin-inject-compiler/ksp/src/main/kotlin/me/tatarka/inject/compiler/ksp/InjectProcessor.kt
```diff
@@ -42,9 +42,17 @@ class InjectProcessor(
...
             resolver.getSymbolsWithAnnotation(COMPONENT.canonicalName).filterIsInstance<KSClassDeclaration>()
-        val deferredClasses = componentSymbols.filterNot { element ->
-            processInject(element, provider, codeGenerator, injectGenerator)
-        }.toList()
+        val deferredClasses = componentSymbols
+            .distinctBy {
+                // Classes keep being processed multiple times and throw a FileAlreadyExistsException in multi round
+                // scenarios. The function above getSymbolsWithAnnotation returns deferred symbols twice for some
+                // reason in the next round.
+                it.qualifiedName
```
evant, line 50: Simple enough, I assume a ksp issue is filed for it? Would be nice to include in the comment here

vRallev: Done https://github.com/google/ksp/issues/1993

evant (approved):

PR #442, opened 2024-09-15
Align and share jvmTarget/JDK versions

ZacSweers:
Was having issues running checks locally because Detekt by default uses the current JDK as the jvm target. This PR seeks to align all these via declaration in the version catalog as well as opportunistically modernize the java configuration too.

### DIFF
--- a/build-logic/src/main/kotlin/kotlin-inject.detekt.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlin-inject.detekt.gradle.kts
@@ -16,6 +16,7 @@ dependencies {
 }
 
 tasks.withType<Detekt>().configureEach {
+    jvmTarget = libs.versions.jvmTarget.get()
     parallel = true
     buildUponDefaultConfig = true
     config.from(file(rootProject.projectDir.resolve(".static/detekt-config.yml")))
--- a/build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
@@ -7,11 +7,16 @@ plugins {
 
 val libs = the<LibrariesForLibs>()
 
-kotlin.compilerOptions.jvmTarget = JvmTarget.JVM_11
+kotlin.compilerOptions.jvmTarget = libs.versions.jvmTarget.map(JvmTarget::fromTarget)
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
+    toolchain {
+        languageVersion.set(libs.versions.jdk.map(JavaLanguageVersion::of))
+    }
+}
+
+tasks.withType<JavaCompile>().configureEach {
+    options.release.set(libs.versions.jvmTarget.map(String::toInt))
 }
 
 // Ensure xml test reports are generated
--- a/build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
+++ b/build-logic/src/main/kotlin/kotlin-inject.multiplatform.gradle.kts
@@ -1,8 +1,6 @@
 import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinNativeTargetWithTests
 import org.jetbrains.kotlin.gradle.targets.js.dsl.ExperimentalWasmDsl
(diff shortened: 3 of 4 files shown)


build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
```diff
@@ -7,11 +7,16 @@ plugins {
...
 
-kotlin.compilerOptions.jvmTarget = JvmTarget.JVM_11
+kotlin.compilerOptions.jvmTarget = libs.versions.jvmTarget.map(JvmTarget::fromTarget)
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_11
-    targetCompatibility = JavaVersion.VERSION_11
+    toolchain {
+        languageVersion.set(libs.versions.jdk.map(JavaLanguageVersion::of))
+    }
```
eygraber, line 15: As per [Jake](https://kotlinlang.slack.com/archives/C3PQML5NU/p1689790193010619?thread_ts=1689789822.225749&cid=C3PQML5NU) regarding `toolchain`:

> A toolchain declaration will force the presence of JDK 17 rather than allowing any JDK 17 or newer to build your project. The Java compiler is designed to cross-compiler older releases. I would not recommend forcing old JDKs as there really should be no reason to use anything but the latest locally (or maybe N-1 while waiting for Gradle support).

Is setting Kotlin and detekt's jvmTarget not enough?

build-logic/src/main/kotlin/kotlin-inject.detekt.gradle.kts
```diff
@@ -16,6 +16,7 @@ dependencies {
 }
 
 tasks.withType<Detekt>().configureEach {
+    jvmTarget = libs.versions.jvmTarget.get()
```
eygraber, line 19: Awesome, this has been a PITA

build-logic/src/main/kotlin/kotlin-inject.jvm.gradle.kts
evant: I was going to link to https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/ as well. I wonder if there's a way to just target this to detekt since that's where the actual issue is.

I've been running locally with java 21 and it's been working fine. I think that would be a better target than 17 if we need to use a specific jdk.

ZacSweers: Tbh I disagree with Jake! At least until Gradle (ever) adds support for a minimum jdk version. I find it eternally annoying to checkout gradle projects that have undefined tooling requirements and manifest with odd/surprising errors. I can limit this PR to just the jvmTarget in detekt if you really prefer though.

eygraber: What about checking the jdk version, and applying a toolchain if it's below the minimum?

ZacSweers: my [LOGAF](https://blog.danlew.net/2020/04/15/the-logaf-scale/) is not high enough to implement that for this PR :)

evant: > I find it eternally annoying to checkout gradle projects that have undefined tooling requirements and manifest with odd/surprising errors.

Fair enough, I think a toolchain set to the latest version of java that doesn't break everything would be an acceptable route

ZacSweers: sounds good 👍. Does that mean you want to raise this to 22?

evant: If 22 works then yep, would also recommend updating https://github.com/evant/kotlin-inject/blob/fe512838343f1234ebbdf87d7c73b94ea3bcad1c/.circleci/config.yml#L54

evant:
bah, circleci [hasn't released jdk 22+ images](https://github.com/CircleCI-Public/cimg-openjdk/issues/205) and the macos runner updated from 17 to 22. So it'd be a real pain to match versions. I think I'm going to merge this without the toolchain support for now since forcing 17 seems too much of a downgrade over the status quo. May revisit in the future.

ZacSweers:
what about just using 21?

evant:
There's not a version of the macos runner that supports 21, they updated from 17 to 22

evant:
Merged

Issue #444, opened 2024-09-27
Repeated bindings from Parent components aren't validated


hectorups:
This code fails as expected because we cannot provide the same binding (a String in this case) multiple times:

```kotlin
@Component
abstract class Example {
    @get:Provides val string1: String = "String1"
    @get:Provides val string2: String = "String2"
}
```

However, in this example the same binding is provided 3 times:

```Kotlin
interface Parent1 {
    val string1: String
}

interface Parent2 {
    val string2: String
}

@Component
abstract class Child(@Component val p2: Parent2, @Component val p1: Parent1) {
    @get:Provides val string3: String = "child"
    abstract val foo: Foo
}

@Inject
class Foo(val string: String)
```

Here is the generated code. It seems, the binding that is used is simply the first one (from Parent2):

```Kotlin
public class InjectChild(
  p2: Parent2,
  p1: Parent1,
) : Child(p2, p1) {
  override val foo: Foo
    get() = Foo(
      string = p2.string2
    )
}
```

shouldn't this code fail for the same reasons than the first example?

PR #445, opened 2024-10-05
check for duplicate bindings between child and parent components

hectorups:
Fixes issue #444

### DIFF
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
@@ -152,6 +152,24 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
 
             val constructor = astClass.primaryConstructor
             if (constructor != null) {
+                fun Result.getAllTypesAndMethods(): Map<TypeKey, AstMember> =
+                    types.mapValues { it.value.method } +
+                        providerTypes.mapValues { it.value.method }
+
+                fun Result.getContainerTypesAndMethods(): Map<TypeKey, AstMember> =
+                    containerTypes.mapKeys { it.key.containerTypeKey(provider) }
+                        .mapValues { it.value.first().method }
+
+                val childAndParentsTypes = mutableMapOf<TypeKey, AstMember>()
+                val childAndParentsContainerTypes = mutableMapOf<TypeKey, AstMember>()
+                val needsToCheckDuplicateTypesWithParents = !accessor.isNotEmpty()
+
+                if (needsToCheckDuplicateTypesWithParents) {
+                    // Start adding the children types
+                    childAndParentsTypes.putAll(getAllTypesAndMethods())
+                    childAndParentsContainerTypes.putAll(getContainerTypesAndMethods())
+                }
+
                 for (parameter in constructor.parameters) {
                     if (parameter.isComponent()) {
                         val elemAstClass = parameter.type.toAstClass()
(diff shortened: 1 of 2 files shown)


kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
```diff
@@ -164,6 +164,11 @@ class TypeCollector(private val provider: AstProvider, private val options: Opti
                             accessor = accessor + parameter.name,
                             typeInfo = elemTypeInfo
                         )
+
+                        // check for duplicates from previously collected Results
+                        for (parent in parents.minus(parentResult).plus(this)) {
+                            checkDuplicateTypesBetweenResults(provider, parentResult, parent)
+                        }
```
evant, line 171: Hm, I don't love that this is n^2. Wondering if it would be better to check once outside of the loop once all the parents have been collected?

hectorups: thanks for the feedback, I'll change that

hectorups:
@evant one of the integration test is failing after the changes to compile:

```kotlin
@Component abstract class ParentComponent {
    abstract val parentNamedFoo: NamedFoo

    @Provides
    protected fun foo() = NamedFoo("parent")

    val Foo.binds: IFoo
        @Provides get() = this
}

@Component abstract class SimpleChildComponent1(@Component val parent: ParentComponent) {
    abstract val namedFoo: NamedFoo

    val BarImpl.binds: IBar
        @Provides get() = this

    abstract val foo: IFoo
}
```

In this example the binding `NamedFoo` is provided by both the parent and child components, creating a situation we're trying to avoid. Currently, a type inside `SimpleChildComponent1` that requires `NamedFoo` will receive the one from the parent. In my opinion, this is arbitrary and likely unwanted, especially in large projects with complex dependency graphs. We should enforce having only one type binding per dependency graph. If multiple bindings of the same type are necessary, qualifiers, type-aliases, or multibindings can be used.

evant:
I don't see see how the type is provided multiple times in that example? Both should be getting an instance from `@Provides protected fun foo()` right?

hectorups:
> I don't see see how the type is provided multiple times in that example? Both should be getting an instance from `@Provides protected fun foo()` right?

👍 my mental model of Dagger, where Components don't use @Provides, caused my confusion.

kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeCollector.kt
hectorups: I haven't yet figured out a way to avoid the O(n^2) complexity. Perhaps memoization could be a solution. Realistically, how many parents can a component have before this is an issue?

One improvement I made is avoiding repetitive checks in recursive calls, which was an issue with deep hierarchies. By adding the check `!accessor.isNotEmpty()`, I aim to ensure checks are performed only once per component.

```kotlin
// When accessor is empty, check for duplicates from previously collected Results
if (!accessor.isNotEmpty()) {
    for (parent in parents.minus(parentResult).plus(this)) {
        checkDuplicateTypesBetweenResults(provider, parentResult, parent)
    }
}
```

evant: You are already collecting into a hashmap to look for duplicates, if you add the previous results each time you should only have to loop once.

evant (approved):

PR #452, opened 2024-10-31
Upgrade KotlinPoet to the latest version

vRallev:
Upgrade KotlinPoet to `2.0.0`, which comes with fixes for line wrapping. kotlin-inject ran into these edge cases for larger components where wrong line wrapping made the code not compile. The latest version of KotlinPoet fixes these issues.

KotlinPoet depends on Kotlin `2.0.10`, therefore upgrade Kotlin and KSP at the same time.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,8 +1,8 @@
 [versions]
 kotlin-inject = "0.7.3-SNAPSHOT"
-kotlin = "2.0.0"
-ksp = "2.0.0-1.0.22"
-kotlinpoet = "1.16.0"
+kotlin = "2.0.10"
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@
 /.idea/*
 !/.idea/codeStyles
 /.kotlin
+local.properties
```
evant, line 7: This isn't an android project so this isn't needed

vRallev: Fair, I use Android Studio for a few KMP projects and it gets generated by default, which is annoying. I don't think excluding this file has a downside, but I'll revert.

evant (approved):

PR #460, opened 2025-01-02
Fix #459: Suppress "Redundant visibility modifier" Warnings in Generated Code

### LINKED ISSUE #459: Suppress Warnings in Generated Code with `@file:Suppress("REDUNDANT_VISIBILITY_MODIFIER")`
When using Kotlin Inject in a project configured with stricter compiler options, such as:

```kotlin
tasks.withType<KotlinCompilationTask<*>>().configureEach {
    compilerOptions {
        extraWarnings.set(true)
    }
}
```

The following warning is generated for Kotlin Inject-generated code:

```
w: file:///path/to/generated/file.kt:line:column Redundant visibility modifier.
```

This warning appears because the generated code includes unnecessary visibility modifiers, which the stricter compiler options flag as redundant.

---

#### Proposed Solution

Add the annotation `@file:Suppress("REDUNDANT_VISIBILITY_MODIFIER")` at the top of each generated file to suppress this specific warning in generated code. 

This change would prevent these warnings from appearing when stricter compiler options like `extraWarnings.set(true)` are enabled, while not impacting other aspects of the generated code or its functionality.

---

#### Example: Generated Code with Suppression

Here’s an example of how the annotation could be included in a generated file:

```kotlin
@file:Suppress("REDUNDANT_VISIBILITY_MODIFIER")

package com.example.generated

public class InjectAppComponent {
    // Generated code here...
}
```

---

#### Benefits
- Keeps the project clean and free of compiler warnings, especially in projects using stricter compiler configurations.
- Does not require developers to modify generated files manually or disable the `extraWarnings` setting for the entire project.

This small change will improve developer experience by ensuring Kotlin Inject-generated code does not cause unnecessary warnings in projects with stricter compiler settings.

leinardi:
Fix for #459

image

image

### DIFF
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/InjectGenerator.kt
@@ -72,6 +72,7 @@ class InjectGenerator(
         val createFunction = createGenerator.create(astClass, constructor, injectComponent, classOptIn)
 
         return FileSpec.builder(astClass.packageName, injectName).apply {
+            addAnnotation(createSuppressAnnotation("REDUNDANT_VISIBILITY_MODIFIER"))
             astClass.containingFile?.optInAnnotation()?.let { addAnnotation(it) }
             createFunction.forEach { addFunction(it) }
             addType(injectComponent)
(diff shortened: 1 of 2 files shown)


evant:
Going to wait for https://github.com/evant/kotlin-inject/pull/350 to be merged first so that a test can be written using -Wextra, ex:

```kotlin
    @ParameterizedTest
    @EnumSource(Target::class)
    fun does_not_warn_by_default(target: Target) {
        val projectCompiler = ProjectCompiler(target, workingDir)

        assertThat(
            projectCompiler.source(
                "MyComponent.kt",
                """

                @Component abstract class MyComponent {
                }
                """.trimIndent()
            ).compile()
        ).warnings().isEmpty()
    }
```

evant:
Yep was planning on adding it eventually if you didn't get to it.

PR #462, opened 2025-01-10
Support type qualifiers in type arguments

### LINKED ISSUE #409: Support qualifieres in multibinds map keys and set elements.

sergeshustoff:
Added support for qualifiers in genric arguments, had to separate type qualifier from member qualifiers, but managed to keep equality between definitions `val qwe: @Qualifier String` and `@get:Qualifier val qwe: String`.
Type qualifier is not equal to member qualifier for `@IntoSet` and `@IntoMap` though, as those have wrapper types

Fixes #409


sergeshustoff:
One of the problems is that type qualifier, even directly on returned type, doesn't always mean exactly the same thing as qualifier on provider function/property for `@IntoSet` and `@IntoMap`. Because of that I had to keep function/property qualifier separate from resulting qualifier in TypeKey, but merge them together with type qualifier for `equals` check.

I think if we stop considering `val qwe: @Qualifier String` and `@get:Qualifier val qwe: String` to be equal we could keep only member qualifier in TypeKey and handle all type qualifiers only in `equals`. That would make code much simpler.

@evant wdyt?

Issue #249, opened 2023-02-02
Better handle empty multiple bindings.


evant:
You may want to declare a multi-binding but provide no values. This is useful for cases where you want to 'extend' the set provided. Right now this is not possible because you have to declare at least one `@IntoMap` or `@IntoSet` in order for it to not complain that it can't provide the value. [Dagger](https://dagger.dev/dev-guide/multibindings.html#declaring-multibindings) has 2 ways around this:

Add `@Multibinds` to the provides

```kotlin
@Multibinds abstract fun aSet(): Set<Foo>
```

Provide an empty collection with `@ElementsIntoSet`

```kotlin
@Provides @ElementsIntoSet
fun primeEmptyFooSet(): Set<Foo> = emptySet()
```

evant:
I'm personally not a fan of having 2 different ways to do the same thing. I'm leaning towards only providing the second as it can be useful in it's own right to group a bunch of multi-binds together into a single method. A way we could do this is with an option to the `@IntoSet` and `@IntoMap` annotations to control it's return type. Ex:

```kotlin
@Provides @IntoSet(multiple=true) fun defaultSet(): Set<Foo> = setOf(Foo("one"), Foo("two"))
@Provides @IntoMap(multiple=true) fun defaultMap(): Map<String, Foo> = mapOf("one" to Foo1(), "two" to Foo2())
```

dave08:
Then it wouldn't be just usable as a default empty set provider, but rather a way to provide multiple `Foo` per `@Provides @IntoSet(multiple=true) `  function? There could potentially be more than one of those and they'll be unioned?

evant:
Yep, exactly

rharter:
Is there a particular preference for overloading `@Into[Set|Map]` vs. having a separate `@ElementsInto...` annotation? I agree with having a single approach, but FWIW I prefer the latter approach since it's a bit more explicit and seems harder to forget to override a property.

evant:
It felt simpler in the moment, but I'm not particularly attached to it, will think about it when implementing.

mrmans0n:
Is it still planned for a 0.7.x ? I am doing nasty workarounds for the time being but official support would be amazing!

evant:
Yes this is still planned, shouldn't be that hard actually if someone wants to contribute!

PR #467, opened 2025-02-19
Add support for IntoSet/Map(multiple = true)

scottjasso:
Fixes https://github.com/evant/kotlin-inject/issues/249

Initial stab at implementing. Haven't added significant tests or updated any docs.

A few notes:

### IntoSet
This changes the `IntoSet` implementation from `setOf()` to something like: 
```kotlin
buildSet(3) {
  add(itemOne())
  add(itemTwo)
  addAll(itemThreeAndFour())
}
```
* `3` here is just the number of providers -- we can't know how many items a `multiple = true` could provide, so our guess is just 1 item per provider. This is the same as what dagger generates for `@ElementsIntoSet`
* We have to use `add` and `addAll` to properly distinguish between a `multiple = true` value and a `multiple = false` value

### IntoMap
Similarly, this changes the `IntoMap` implementation from `mapOf()` to something like:
```kotlin
buildMap(3) {
  this += pairOne()
  this += pairTwo()
  this += mapOfThreeAndFour()
}
```
* In this case, we can use the `+=` operator since we know the values are either `Pair` or `Map` -- it's unambiguous.

### IntoSet(multiple = True) w/ Lazy or Functions
If you have a binding of `Set<Lazy<String>>` or `Set<() -> String>`, along with an `IntoSet(multiple = True)`, I'm not sure what is expected. Wrapping the provider with `lazy` wouldn't work -- we'd have to actually evaluate the provider anyway. For now I fail fast with an error.

### DIFF
--- a/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
+++ b/kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
@@ -12,6 +12,7 @@ import com.squareup.kotlinpoet.PropertySpec
 import com.squareup.kotlinpoet.TypeName
 import com.squareup.kotlinpoet.TypeSpec
 import com.squareup.kotlinpoet.tags.TypeAliasTag
+import com.squareup.kotlinpoet.withIndent
 
 data class TypeResultGenerator(val options: Options, val implicitAccessor: Accessor = Accessor.Empty) {
 
@@ -68,7 +69,8 @@ data class TypeResultGenerator(val options: Options, val implicitAccessor: Acces
             is TypeResult.Provides -> generate()
             is TypeResult.Scoped -> generate()
             is TypeResult.Constructor -> generate()
-            is TypeResult.Container -> generate()
+            is TypeResult.SetContainer -> generate()
+            is TypeResult.MapContainer -> generate()
             is TypeResult.Function -> generate()
             is TypeResult.NamedFunction -> generate()
             is TypeResult.Object -> generate()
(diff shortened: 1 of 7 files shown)


kotlin-inject-compiler/core/src/main/kotlin/me/tatarka/inject/compiler/TypeResultGenerator.kt
```diff
@@ -245,18 +247,31 @@ data class TypeResultGenerator(val options: Options, val implicitAccessor: Acces
...
         return CodeBlock.builder().apply {
-            add("$creator(")
-            add("\n⇥")
-            args.forEachIndexed { index, arg ->
-                if (index != 0) {
-                    add(",\n")
+            beginControlFlow("buildSet(%L)", args.size)
+            withIndent {
+                for (arg in args) {
+                    if (arg.key.type.isSet()) {
```
scottjasso, line 255: actually I think i'll need a different check here. `isSet` wouldn't work for a `Set<Set<T>>` binding.

evant:
Just want to say thank you for this! Will get to a code review when I get a chance but at a quick glance it looks good!

> IntoSet w/ Lazy or Functions

yeah that's tricky, will have to think about this. Option 1 is the most conservative so I agree that's a good starting point.

scottjasso:
> Curious what dagger does? I believe it has an equivalent feature set.

it actually doesn't allow this behavior: https://dagger.dev/dev-guide/multibindings

> As with any other binding, in addition to depending on a multibound `Set<Foo>`, you can also depend on `Provider<Set<Foo>>` or `Lazy<Set<Foo>>`. You cannot, however, depend on `Set<Provider<Foo>>`.

PR #472, opened 2025-03-18
update readme documentation for assisted injection to cover the assisted factory

plusmobileapps:
I just noticed the use of `@AssistedFactory` in the [sample](https://github.com/evant/kotlin-inject/blob/main/integration-tests/common/src/main/kotlin/me/tatarka/inject/test/Assisted.kt#L27), but didn't see it well documented. So I updated the readme to explain the usage of this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 /.idea/*
 !/.idea/codeStyles
 /.kotlin
+local.properties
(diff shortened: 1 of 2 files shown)


plusmobileapps:
I guess this is in a snapshot build because I don't actually see this on the latest available version in my project.

PR #475, opened 2025-04-17
Update Kotlin and KSP

eygraber:


eygraber:
Same OOM issues with CI. Maybe a `Xmx` bump and/or a larger runner is needed?

droideveloper:
> Same OOM issues with CI. Maybe a `Xmx` bump and/or a larger runner is needed?

would you like to add "-XX:MaxMetaspaceSize=1024m" or "-XX:MaxMetaspaceSize=1G" in "grade.properties" for "org.gradle.jvmargs", so CI implementations need that for their JVM, if that not help might want to bump heap to "6G" or "8G" in there without tear change, I did try this runner command in Mac osx, did not fail, did not have linux setup to give a shot, thanks @eygraber

evant:
the linux ci runner has 8GB of memory so there should be some headroom if you want to bump memory usage

eygraber:
I bumped it to 6G, let's see what happens.

evant:
just merged in some changes that should greatly lower memory usage, rebasing might fix the issues here.

eygraber:
Still broken. I'll dig a little deeper.

evant:
ironically I think bumping it to 6Gb causes it to fail, I think the gradle daemon is taking away memory from the kotlin daemon https://app.circleci.com/pipelines/github/evant/kotlin-inject/1134/workflows/27bdcd96-e1c7-42ae-9431-46ab9490ca84/jobs/1887

eygraber:
Ah I was actually thinking about that when I made the change :see_no_evil:

evant (approved):

PR #477, opened 2025-04-26
improve circleci caching

evant:


eygraber:
Would you be willing to switch to using GitHub Actions? I have a pretty good flow in place for my own libraries that I could implement here if you want. It seems simpler than what's going on with CircleCI.

evant:
I've been using circelci forever so it's the one I'm the most familiar with. It also has no build credit limit unlike github actions.

eygraber:
I'll start putting something together.

[Here's an example of what releases would look like](https://github.com/eygraber/sqldelight-androidx-driver/releases) (customizable via [changelog_config.json](https://github.com/eygraber/sqldelight-androidx-driver/blob/master/changelog_config.json)).

Snapshots are published after PRs are merged, and releases are created and published by manually running the [Publish release workflow in GitHub Actions](https://github.com/eygraber/sqldelight-androidx-driver/actions/workflows/publish_release.yml).

Required secrets:

- `PUSH_PAT`: a GitHub token for a (bot) account that bumps the version after a release is made and automatically tracks changes to `package-lock.json` or `yarn.lock`
- `SONATYPE_NEXUS_USERNAME`
- `SONATYPE_NEXUS_PASSWORD`

Optional secrets:

- `ARTIFACT_SIGNING_PRIVATE_KEY`: a PGP key for signing the artifacts when publishing to maven central
- `GIT_SIGNING_PRIVATE_KEY`: a PGP key for signing the version bump commit and release tag
- `GIT_SIGNING_PRIVATE_KEY_PASSWORD`: password for `GIT_SIGNING_PRIVATE_KEY`

evant:
create https://github.com/evant/kotlin-inject/discussions/486 to move discussion to so it's not lost the in ether

PR #485, opened 2025-04-27
add release instructions

evant:


eygraber (approved):
Sticking with CircleCI?

evant:
release process should be the same regardless

PR #492, opened 2025-05-21
Add GitHub actions

ArthurKun21:
Hello

was thinking of checking kotlin-inject for migration of my android project into KMP and then I saw https://github.com/evant/kotlin-inject/discussions/486 and figured I can do the simple stuff of building initial setup of Github actions and let you take care of the publishing and other complex stuff.

---

Changes made with this PR
- Enabled build whenever there is a push on the `main` branch or in a `pull_request` can also manually building at the actions tab
- Made the concurrency to cancel other build if done not on the main branch
- Setup a matrix so the `macos` and `ubuntu` can run in parallel
- Did not explicitly added the chrome and chrome drivers as it is already on the runner-images
  - https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#browsers-and-drivers
  - https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md#browsers
- Make use of git hash for the version control of github actions for security
- Added ability to ignore when you update a markdown and the gitignore
- `gradle-wrapper-validation` for validating you got true gradle wrapper https://github.com/gradle/actions/blob/main/docs/wrapper-validation.md#gradle-wrapper-validation-action
- `setup-gradle` was added so you can cache the build. By default it will only cache the default branch and other branches will only be a read only to it https://github.com/gradle/actions/blob/main/docs/setup-gradle.md#using-the-cache-read-only

I also made another workflow for lint check. This one will always run on all of the commits in the `pull request` and pushes to `main` branch

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,6 +39,9 @@ jobs:
         with:
           java-version: ${{ env.JAVA_VERSION }}
           distribution: ${{ env.JAVA_DISTRO }}
+
+      - name: Set up gradle
+        uses: gradle/actions/setup-gradle@8379f6a1328ee0e06e2bb424dadb7b159856a326 # v4.4.0
       
       - name: Build ${{ matrix.os }}
         run: ./gradlew assemble
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,72 @@
+name: Build
+
+on:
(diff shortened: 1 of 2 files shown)


evant:
oh thanks, will take a look!
may run these in parallel for a while to see if github actions is more reliable

PR #505, opened 2025-09-17
Update Gradle to 9.1.0

eygraber:


eygraber:
I didn't enable it here, but the configuration cache seems to be working (at least when running `./gradlew build`).

evant (approved):

